PR #1, opened 2018-11-02
Workingbranch2   merging proposal 2+3 pp

PrandiniUniPD:
My proposal for the merging of the demo from group 2 and the demo from the group 3.


frankplus:
I tested the app and noticed that the captured image is rotated counterclockwise by 90°, any idea why?

PrandiniUniPD:
Thanks for testing.
It is a known bug about the app of the group 2 already present in the demo we've shown to the professor. So for now I'm doing only merging between demos and then I will propose new features and I will ask solutions for bugs.
I've proposed this merge to my group but if you and your group like this merge we can make a common demo quickly and then resolve bugs.

PrandiniUniPD:
Maybe the solution is already included in the demo of the group 1.
I think that in the demo of the group 1 is also included the work about the restore of the last captured pic result.
(the group 2 has used a button to see the last result but the professor wants only the button for taking pics and so I've removed it)

frankplus:
A possible solution may already be in our demo too:
```
            //rotate image by 90 deg
            Matrix matrix = new Matrix();
            matrix.postRotate(90); // anti-clockwise by 90 degrees
            bitmap = Bitmap.createBitmap(bitmap , 0, 0, bitmap.getWidth(), bitmap.getHeight(), matrix, true);
```

frankplus:
I corrected the 90 degrees image rotation bug, with the fix our group proposed. We'll later see if group 1's fix is better.

PrandiniUniPD:
Thank you. Is the rest of the app ok for you and your group?

frankplus:
Per il nostro gruppo è ok

frankplus:
Per il nostro gruppo è ok

frankplus (approved):
It's ok for our group

PrandiniUniPD:
Thank you, I'm waiting the agreement from my group mates and then I'll accept this merge and I'll propose this mixed demo also to the others group with a pull request directed to the branch "merge_2+3".

Issue #13, opened 2018-11-04
Rotated photo on some devices


lucamoroz:
On some devices the captured photo is rotated by 90 degrees

GiovanniFurlan:
me too on a redmi 4x      7.1.2

lucamoroz:
Here is proposed a solution which need an URI to get the image orientation and act accordingly:

https://stackoverflow.com/questions/14066038/why-does-an-image-captured-using-camera-intent-gets-rotated-on-some-devices-on-a

PrandiniUniPD:
@GiovanniFurlan has implemented a possible solution for that. You can find it in the branch named CommonDemo.
In my Wileyfox Swift 2 Plus (API level 27 - Android 8.1) it works.

PrandiniUniPD:
It works for my device, but if not for the others there is another guide here https://firebase.google.com/docs/ml-kit/android/recognize-text . This is the official guide for ml-kit, in this page there is an example of the code useful to recognize the orientation of the Image object.

PrandiniUniPD:
I've checked more precisely the solution by @GiovanniFurlan but it's not complete.

PrandiniUniPD:
I've opened a pull request with the possible solution ( #16 )
The branch where I've worked is named "RotationBug"

PrandiniUniPD:
The pull request has been accepted because #20 has a code derived from #16 
In the Common Demo branch there is a version that use the gyroscope, it's not always the best method.
The proposed method is with the use of the Surface orientations.

Nintenking:
Is this issue still active?

PrandiniUniPD:
Last time some people have said that on their smartphones not work the proposed solution (in #16 there are more details).
In #29 there is another version of the CameraActivity, so when that pull request will be accepted we have to assure that this issue is definitely closed, otherwise for some people still the problem and this issue still opened until a solution.

PrandiniUniPD:
On my smartphone with API level 27 this issue is solved with the version contained in the CommonDemo branch.
Anyone has this problem?

Nintenking:
Tested on my smartphone, API level 26. Although I never actually experienced this problem before, I run anyway various tests with both device rotation activated and deactivated; the captured image was always the one expected, never wrongfully rotated.
Not a problem for me

PrandiniUniPD:
Outdated

PR #16, opened 2018-11-05
Rotation bug Resolved in CameraActivity, method imageOrientation ?

PrandiniUniPD:
This is a branch to solve the rotation bug reported in the issue #13 
Test it, it works for my device.

Filter the logs with the keyword "imageOrientation" for more details.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -29,7 +29,7 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="1.8" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" project-jdk-name="1.8" project-jdk-type="JavaSDK">
     <output url="file://$PROJECT_DIR$/build/classes" />
   </component>
   <component name="ProjectType">
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -29,7 +29,7 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" project-jdk-name="1.8" project-jdk-type="JavaSDK">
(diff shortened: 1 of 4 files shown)


lucamoroz:
The solution solved the bug for my Xiaomi mi6, but using a gyroscope may cause some problems when taking a picture with the phone barely parallel to the ground leading to a wrongly rotated image

frankplus:
Seems to be working on Xiaomi mi a2 too.

PrandiniUniPD:
In your opinion, could we accept this solution with the scope of a demo for tomorrow and eventually in the future implement the landscape mode and another way to resolve this bug?

mattia1142866:
ho provato a fare la foto con il cellulare orientato in opposite landscape e si l'ha ruotata, ma di 90 gradi in più [Xiaomi mi A2 Lite]

PrandiniUniPD:
> ho provato a fare la foto con il cellulare orientato in opposite landscape e si l'ha ruotata, ma di 90 gradi in più [Xiaomi mi A2 Lite]

What does the Log filtered by "imageOrientation" say?
Is it the case that has said @lucamoroz ?

> The solution solved the bug for my Xiaomi mi6, but using a gyroscope may cause some problems when taking a picture with the phone barely parallel to the ground leading to a wrongly rotated image

PrandiniUniPD:
There is another solution for this bug, but we should implement and enable the landscape mode for this new one.
It consists to use the Surface orientations.
But I've some problem to change the camera preview. When I rotate the device from the portrait mode, the mode changes in landscape mode but the camera remains with the characteristics of the portrait mode with the difference that the dimension of the preview is stretched to fit the TextureView. 
Any ideas?

mattia1142866:
i've search around internet and a lot of solutions are using the Inflaters. I think is a possible solution for change orientation of the screen and the elements that make up the layout

mattia1142866:
This link talking about orientations using Camera2 API:
https://medium.com/@kenodoggy/solving-image-rotation-on-android-using-camera2-api-7b3ed3518ab6

Issue #18, opened 2018-11-05
The scanned text is not properly ordered


PrandiniUniPD:
The lines of the scanned text is not always printed in the right order.
It often happen when the scanned text is fragmented in multiple lines not so condensed between them.

PrandiniUniPD:
In the branch ScannedTestSorting there is implemented a version of TestExtractor.java that returns the String of the text extracted ordered by the recognized FirebaseVisionText.TextBlocks from top to bottom.
This version don't order the blocks also from left to right, it's not so useful because the pic will be cropped and the focus is only on the ingredients block.
Anyway it is in the ScannedTestSorting branch.

PR #20, opened 2018-11-08
Text extractor to string

PrandiniUniPD:
Proposal to a version of the class TextExtractor compatible with the OCRInterface (without observer).
That version return a String of the text recognized by the ocr.
Maybe this class not resolve the issue #18 and the other open issues.
Watch only the text extractor pieces of code.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -29,7 +29,7 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="1.8" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" project-jdk-name="1.8" project-jdk-type="JavaSDK">
     <output url="file://$PROJECT_DIR$/build/classes" />
   </component>
   <component name="ProjectType">
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -29,7 +29,7 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" project-jdk-name="1.8" project-jdk-type="JavaSDK">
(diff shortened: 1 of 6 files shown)


frankplus (approved):
ok for me

frankplus:
Does this resolve the rotated photo issue too?

PrandiniUniPD:
> Does this resolve the rotated photo issue too?

This branch is derived from the branch used for solving that issue.
Now it uses the gyroscope for trying to solve that bug but in the pull request #16 there are more details than here, so for that bug we could refer to that pull request.

lucamoroz:
Tested, seems good to me

lucamoroz (approved):
Tested and working

GiovanniFurlan:
There is no "No text find, take new pic" so when no text is found there is white screen on text.

PrandiniUniPD:
> There is no "No text find, take new pic" so when no text is found there is white screen on text.

Thank you, I've missed that feature. Now it is in the code.

GiovanniFurlan (approved):

frankplus:
A downside of this change could be that parallelism between multiple ocr is not possible because of `while (!fbText.isSuccessful());`

stefanoromanello (approved):
ok

PrandiniUniPD:
> A downside of this change could be that parallelism between multiple ocr is not possible because of `while (!fbText.isSuccessful());`

For parallelism could it work creating an instance of TextExtractor for every Thread that needs an ocr analyzing?

Anyway that statement is a problem because it isn't resource friendly.
I've tried with multiple thread and semaphores but not working properly, maybe i've missed something.
So I've put a ToDo there to indicate a required optimization.
Any ideas for substituting it with a resource friendly statement?

PrandiniUniPD:
Four reviewers are agree to me for approving this pull request, the code of this pull request is useful for the tests we have to do for tomorrow and I've contacted @EliaBedin for speak about the usefully of this pull request and the inefficiency presents in this code (@EliaBedin and I have tested some alternatives).
So we can conclude that we can accept this pull request but we have to introduce a note that speak about the inefficiency, I'll put it in a new issue.

Issue #22, opened 2018-11-09
Too much time to elaborate photos


GiovanniFurlan:
The app takes more then 20 seconds to elaborate the taken pics. This happens because there is the write on storage of the image. If we pass the text and the image file path through preferences we can speed up with more comfort for future test and users.

PrandiniUniPD:
Work approved in #25

Issue #12, opened 2018-11-04
Taking photos lag


lucamoroz:
The effective taken photo is not the one displayed on the screen when the button is pressed, but it's a photo taken a considerable amount of time later

Fisher4537:
How much is the time gap between the pressure of the botton and the take of the photo in your configuration? I tried only in a pixel API 25 and the gap is 1.25s.

lucamoroz:
It takes about 5 seconds to take the effective picture plus at least more 5 seconds to extract the text

PrandiniUniPD:
Resolved in #29

PR #23, opened 2018-11-09
Photo testing better performance

GiovanniFurlan:
Strong speed up on text recognition, deleting the data saving on storage. 
Now the app saves the image path and the text on preferences, all preferences are related to the same manager "prefs" and default value are set on null.


PrandiniUniPD (approved):
The app is very fast and it seems to work well.

PrandiniUniPD:
( This is a possible solution for the issue #22 )

lucamoroz (approved):
The app is faster on the text recognition but #12 is still a problem for me

PrandiniUniPD:
The last push resolves the new conflicts generated by accepting the pull request #20 .
For an optimal performace is suggested to use AsyncTask to do in background the work of TextEctractor from the ResultActivity.

mattia1142866 (approved):

PR #25, opened 2018-11-14
TextExtractor inefficiency solution - performace improved

PrandiniUniPD:
This pull request solve the inefficiency introduced for respecting the OCRInterface ( #24 ) and there is some improvement of the performance.
Also there is the actual code for testing the database but it will be improved.

PR #26, opened 2018-11-19
Bug fix backward compatibility

Fisher4537:
### Proposal to solve the app crash with API level lower than 24
Added drawable/roundedbutton.xlm which is a copy of drawable-v24/roundedbutton.xml
### Executed Tests
* API 21 (Android version 5.0.1), physical device, Samsung Galaxy S4 GT-I9505
* API 28, emulator, Pixel


Fisher4537:
roundedbutton.xml is no longer in use. Issue #27 is resolved after pull request #25.

PR #28, opened 2018-11-19
removed drawable-v24/roundedbutton.xml: no longer in use

Fisher4537:
Removed drawable-v24 roundedbutton is no longer in use.


PrandiniUniPD:
The last commit works for a virtual machine with API 22, there were some statement in the wrong places.
(I've not touched any xml and now it seems to work, I'm not sure that was the problem, but I have only virtual machines with API level < 27 for testing, so some tests with at least a physical machine is required for approving this pull request)

stefanoromanello:
In #29 I've also removed roundedbutton.xml and changed the camera. This pull request is no longer needed

PrandiniUniPD:
> In #29 I've also removed roundedbutton.xml and changed the camera. This pull request is no longer needed

A pull request with this changes is already included in CommonDemo, so we could close this pull request

PR #29, opened 2018-11-21
Fixed Camera lag

stefanoromanello:
Implemented a new camera. Camera2 completely removed.


frankplus:
This looks like a significant change, can you explain more what you did? why have you opted for camera-kit instead of camera2?

By the way it's not working on my phone, when I press the take picture button nothing happens.

PrandiniUniPD:
I've tested the app with a virtual machine (API level 28 and API level 22) and with my physical smartphone (API level 27).
With the virtual machine it's not work, it's not clear why, is it possible that is a problem of the vm?
With the physical smartphone it seems to work, the pic saved is the pic that I've took by clicking the button.
I'm interested to this kit, it seems that is less complex than Camera2 but I quote the questions that has written @frankplus for knowing more details.

> This looks like a significant change, can you explain more what you did? why have you opted for camera-kit instead of camera2?
> 
> By the way it's not working on my phone, when I press the take picture button nothing happens.

Nintenking:
I just tested your branch, both on an emulator (API 27) and on my device work just fine, didn't lag almost at all.
On a side note I perceive the text recognition is significantly better now, so far this is a great improvement for me

Nintenking:
I didn't have any issue using it in a vm, perhaps it's a problem on your pc configuration.

I also would like to understand why the change on the implemented camera -it's not a problem for me, I think this is for the better-

> With the virtual machine it's not work, it's not clear why, is it possible that is a problem of the vm?
> I'm interested to this kit, it seems that is less complex than Camera2 but I quote the questions that has written @frankplus for knowing more details.

stefanoromanello:
> This looks like a significant change, can you explain more what you did? why have you opted for camera-kit instead of camera2?
> 
> By the way it's not working on my phone, when I press the take picture button nothing happens.

Strange, for now it seems that its a problem only for you. Try to re-pull the branch from the last commit. 
I've removed camera2 because its really complicate to use and because it was the problem of the lag.

frankplus:
I solved it, It was a problem on my phone. It works very well now

leopardo-rossi-developer:
I tried the new camera and the first time it didn't work because there were some errors. The first one was that sometimes the "orientationResult" variable was null so I added an if to avoid this point of failure. The second problem was in AsyncLoad class. Indeed here, when the text was extracted from the OCR, the class tried to set the text on a label on the main thread but it couldn't be reach. So I added some lines of code to avoid this problem. I've done a push with these things fixed and now all works fine! Take a look and tell me if something it's not clear.

stefanoromanello:
For me all the changes made until now are fine. Please approve this merge because is a big part of the project.

Nintenking (approved):
Overall I find it better performing than Camera2, I'm fine with this changes

leopardo-rossi-developer (approved):

PrandiniUniPD (approved):
These changes make the CameraActivity less complex than before and the performances are improved.
Maybe we should assure that we can trust to this kit because it's a beta version.
We need a common style document for the code.  For now could we use the Google Code Style? https://source.android.com/setup/contribute/code-style

It would be useful that at least one people from every group tests this version of the app and makes his review about it because then we can focus to other parts of the app.

frankplus (approved):

ghost:
Tested this new changes:
- there is less lag _(still present if the phone is not high-end or midrange but the problem has been mitigated at least)_ when capturing a photo;
- text recognition seems a little faster.

The only things that's missing is autofocus (or tap to focus), but still this is a solid update of the past camera activity for CommonDemo, approved for me.

PR #34, opened 2018-11-26
Barcode reader

Fisher4537:
Implemented a first version of ML Kit barcode reader.
Advices are encouraged.


PrandiniUniPD:
I've tested this version of the app and it seems that recognizes the barcodes correctly, maybe a title above the text extracted and another above the extracted barcode would be useful to visually differentiate the two types of extraction.
But would you add this feature now? Or only in the case this feature is requested by the app?

Fisher4537:
Thanks for your review and your tips! I've added the string "Barcode:" in the barcode TextView

PR #35, opened 2018-11-27
Brightness recognition

BalzPit:
Added a method "IMG_Brightness" that determines wether or not the image is too dark or bright.
Before calling the intent at the end of TakePhoto in CameraActivity the method is called and if it detects over or underexposure a Toast will pop up notifying the user


PrandiniUniPD (commented):
I've tested this app. I've received the toast in the ResultActivity after some time because the app waits the result of the recognition before continuing with the ResultActivity.
Is it a good idea running this recognition to another thread in the ResultActivity?
If not, is it a good idea running the recognition in an acynctask to notify the user the app are working to him?
For too bright pictures I've test a photo of a light is it right? (toast not received)
For too dark picture I've test a photo of a no light environment is it right? (toast received)

photo_2018-11-27_11-16-16
photo_2018-11-27_11-16-10

frankplus (approved):
Looks good to me

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -156,6 +163,17 @@ public void onImage(CameraKitView cameraKitView, final byte[] photo) {
                     edit.putString("imagePath", filePath.trim());
                     edit.apply();
 
+                    // analyze the brightness   @author Balzan Pietro
+                    final Bitmap bitmapImage2 = bitmapImage;
+                    new Thread(new Runnable() {      // run the brightness recognition on a new thread to improve performance
```
PrandiniUniPD: Is it really useful running these statements in a new Thread considered that the code of the BrightnessRecognition is in another java class and so in another Thread?

app/src/main/java/unipd/se18/ocrcamera/BrightnessRecognition.java
```diff
@@ -0,0 +1,93 @@
...
+     * @param upperBound the toast will be launched if brightness exceeds this value.
+     * @param lowerBound the toast will be launched if brightness is lower than this value.
+     * @param pixelSkip how many pixels to skip each checked pixel. Higher values result in better
+     *                     performance, but a more rough estimate. When pixelSpacing = 1, the method
+     *                     actually calculates the real average brightness, not an estimate.
+     *                     Do not use values for pixelSpacing that are smaller than 1.
+     * @author Pietro Balzan - some changes by Francesco Pham
+     */
+
+    public void imgBrightness(int upperBound, int lowerBound, int pixelSkip){
```
PrandiniUniPD: Could be useful if this method returns an int for a possible future automatic processing of the picture? For example -1 too dark, 0 ok, 1 too bright?
Is really useful setting externally these parameters? Or they could be calculated automatically or anyway internally to this method?

app/src/main/java/unipd/se18/ocrcamera/BrightnessRecognition.java
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Constructor of a BrightnessRecognition object containing a Bitmap
+     * @param bmp the bitmap of the image to calculate the brightness of
+     * @param cntxt context of the activity that instances an object of this class
+     * @param hndlr handler for communication with ui thread
+     * @modify image with the new bitmap
+     * @author Pietro Balzan
+     */
+
+    public BrightnessRecognition(Bitmap bmp, Context cntxt, Handler hndlr){
```
PrandiniUniPD: Could have sense if this class contains only static methods without a constructor?

PrandiniUniPD (commented):
I prefer this version with the processing in another class, good work.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
BalzPit: Maybe i coded it wrong but when I tried just running the statements like you said, the performance wasn't improved: it still waited for the Recognition to end before launching the ResultActivity with the intent. It was like there wasn't a new automatic thread, so I made it manually.

app/src/main/java/unipd/se18/ocrcamera/BrightnessRecognition.java
BalzPit: good idea, I will add that feature

app/src/main/java/unipd/se18/ocrcamera/BrightnessRecognition.java
BalzPit: It makes sense

PrandiniUniPD (approved):

Nintenking (approved):
Works fine for me, could you now do a Pull request to the branch "ImagePorcessing_module"? We're making a dedicated module and putting there all the Image processing features

frankplus:
Just for the record, I tested the time of execution of imgBrightness(), on my phone (Mi A2) with pixelSkip=3 it takes about 130 ms, it is ok to me.

Fisher4537:
Why is this pull request already open? @BalzPit is it alrready active? We discussed about the class which should countain all the methods of image manipulation! Do you change your mind?

Nintenking:
This pull request was made before our idea of the Image Processing module, and we already merged this branch in the one with the module. He surely just forgot to close this pull request, so I'm doing it now.

PR #38, opened 2018-11-29
CodeReview of 15 from 1

stefanoromanello:


Nintenking:
I'm not as experienced as you in coding so I'll do my best to point out what I think could be improved!

Nintenking:
I would like to point out one thing I think we agreed on the code style: when writing a method you sometimes use the format

`public method()`
`{`
   ` //code`
`}`

but we thought that we all should try to do it like this


`public method(){`
    `//code`
`}`


Because I think using a standard way to put the braces should help to avoid a number of conflicts while trying to merge two branches

stefanoromanello:
Thanks for the answer. your suggestions is correct and I will take care on the code that I will write but for this I dont think this is relevant so please approve this merge if you don't find any other problems. Thanks

Nintenking:
Perfect, just one little thing: if for the next time you could comment the code a little more, it would be **gold** for a newbie like me!
I'll close the pull request now since I didn't find anything else to point out. Good job!

PR #39, opened 2018-11-29
CodeReview of 15 from 7

stefanoromanello:


app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,250 @@
...
+/**
+ * The Activity useful for making photos
+ */
+public class CameraActivity extends AppCompatActivity {
+
+    private CameraKitView cameraKitView;
+    private static String orientationResult;
+
+    /**
+     * onCreate method of the Android Activity Lifecycle
```
GiovanniFurlan: Could be better to add @modify the shared preferences on button click?

GiovanniFurlan, line 34: Could be better to add @modify the shared preferences on button click?
Adding a link where to find the library would be really nice.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,250 @@
...
+            {
+                float[] values = event.values;
+                int orientation = ORIENTATION_UNKNOWN;
+                float X = -values[_DATA_X];
+                float Y = -values[_DATA_Y];
+                float Z = -values[_DATA_Z];
+                float magnitude = X*X + Y*Y;
+                // Don't trust the angle if the magnitude is small compared to the y value
+                if (magnitude * 4 >= Z*Z) {
+                    float OneEightyOverPi = 57.29577957855f;
```
GiovanniFurlan: Could be better to add final on OneEightyOverPi or define it at the beginning of the method

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,250 @@
...
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_camera);
+
+        cameraKitView = findViewById(R.id.cameraKitView);
+
+        //Load sensor for understand the orientation of the phone
+        SensorManager sensorManager = (SensorManager) getSystemService(Context.SENSOR_SERVICE);
+        sensorManager.registerListener(new SensorEventListener() {
+            public int mOrientationDeg; //last rotation in degrees
```
GiovanniFurlan, line 49: I have a problem understanding how the sensor works. 
Are _DATA_X , Y and Z the axes? 
A comment explaining the working idea would make the method more clear

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,250 @@
...
+                        Log.d("Sensor", "LL"); //LandscapeLeft
+                        orientationResult="LL";
+                        takePicButton.setRotation(90);
+                    }
+
+                }
+            }
+
+            @Override
+            public void onAccuracyChanged(Sensor sensor, int accuracy) {
```
GiovanniFurlan, line 118: I didn't understand why is the method onAccuracyChanged  void?
Could be good to add a comment to explain that

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,250 @@
...
+        File outputDir = context.getCacheDir();
+        File imageFile = new File(outputDir, name + ".jpg");
+
+        OutputStream os;
+        try {
+            os = new FileOutputStream(imageFile);
+            bitmap.compress(Bitmap.CompressFormat.JPEG, 100, os);
+            os.flush();
+            os.close();
+        } catch (Exception e) {
```
GiovanniFurlan, line 231: Should be better to not catch every Exception but only the one referred to OutputStream

GiovanniFurlan (changes requested):
Hi, I've reviewed the class.
I didn't completely understand part of your code, so please be patient.
The code is good and i've learned some tips. Thanks.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
stefanoromanello: Thanks for the suggestion but at first glance this could be a great observation but in this case I have to leave as it is. The problem is that I could change "Exception" to "FileNotFoundException" but os.flush(); and os.close(); require "Exception" as catch so if I want to include them into the try catch I have to leave "Exception". Anyway "Exception" also trigger "FileNotFoundException"

stefanoromanello:
Ive added your suggestions. Please tell me what you think

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
GiovanniFurlan: ok, thank for the explanation

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
GiovanniFurlan: i add final or private to the variable to avoid other method to modify it

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -27,10 +29,11 @@
 
     private CameraKitView cameraKitView;
     private static String orientationResult;
-
+    float OneEightyOverPi = 57.29577957855f;
```
GiovanniFurlan: Could be better add private or final to the variable to avoid other method to modify the value

GiovanniFurlan (changes requested):

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
stefanoromanello: What, I've changed it before the push, was 
`float final OneEightyOverPi = 57.29577957855f;`
Sorry I meant to set it as final... Going to push now

stefanoromanello:
Fixed final variable

GiovanniFurlan:
It's all ok now

GiovanniFurlan (approved):

PR #42, opened 2018-11-30
CodeReview of 9 from 1

leopardo-rossi-developer:
The method's for the code review are: takePhoto() and tempFileImage().

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
@@ -1,0 +1,316 @@
+package unipd.se18.ocrcamera;
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.Intent;
+import android.content.SharedPreferences;
+import android.graphics.Bitmap;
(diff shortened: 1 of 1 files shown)


Nintenking:
I didn't really find anything bad in the code structure worth commenting, seems to do just fine what it wants to do.
The code is very well commented, easy to understand what you are trying to do in every line, quite _nice_ while reading.
The only thing I would like to point out is one thing I think we agreed on the code style: when writing a method you use the format

`public method()`
`{`
   ` //code`
`}`

but we thought that we all should try to do it like this


`public method(){`
    `//code`
`}`


Because using a standard way to do put the braces should help to avoid a number of conflicts while trying to merge two branches!

Besides that, very well done!

leopardo-rossi-developer:
Thank you for your comment and review! I think that braces aren't very important for code quality because its meaning is exactly the same

Nintenking:
Of course, they are the same, it should be done only so we all had the same standard.
Anyway, this is all correct for me, I'm closing this pull request. Well done!

PR #43, opened 2018-11-30
CodeReview of 8 from 0

PietroPrandini:
From the code I've written, for this code review I've chosen to propose to you only the TextExtractor.java because it has a core functionality for the app.
There are many comments and links for supporting the review.
I hope you can review easily this code, if not, I'm here.
Have a good review.

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
@@ -1,0 +1,240 @@
+package unipd.se18.ocrcamera;
+
+import android.graphics.Bitmap;
+import android.util.Log;
+import com.google.android.gms.tasks.OnSuccessListener;
+import com.google.android.gms.tasks.Task;
+import com.google.firebase.ml.vision.FirebaseVision;
+import com.google.firebase.ml.vision.common.FirebaseVisionImage;
+import com.google.firebase.ml.vision.text.FirebaseVisionText;
+import com.google.firebase.ml.vision.text.FirebaseVisionTextRecognizer;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.CountDownLatch;
+
+/**
+ * Class that implements the common OCR interface for retrieving text from a Bitmap image.
+ * This class uses ml-kit provided by Firebase.
+ * More details at: {@link FirebaseVisionText}.
+ * @see <a href="https://firebase.google.com/docs/ml-kit/android/recognize-text">
+ *     ml-kit, recognize text</a>
+ * @author Pietro Prandini (g2)
+ */
+public class TextExtractor implements OCRInterface {
+    /**
+     * String used for the logs of this class.
+     */
+    private final String TAG = "TextExtractor";
+
+    /*
+    The next four int are used for recognizing the position of the FirebaseVisionText Objects.
+    They could be useful for sorting the blocks or for an automatic recognition
+    of the ingredients text block.
+     */
+
+    /**
+     * Index of the point on the top-left position.
+     * More details at:
+     * {@link FirebaseVisionText.TextBlock#getCornerPoints()},
+     * {@link FirebaseVisionText.Line#getCornerPoints()}
+     * or {@link FirebaseVisionText.Element#getCornerPoints()}.
+     */
+    private final int TOP_LEFT = 0;
+
+    /**
+     * Index of the point on the top-right position.
+     * More details at:
+     * {@link FirebaseVisionText.TextBlock#getCornerPoints()},
+     * {@link FirebaseVisionText.Line#getCornerPoints()}
+     * or {@link FirebaseVisionText.Element#getCornerPoints()}.
+     */
+    private final int TOP_RIGHT = 1;
+
+    /**
+     * Index of the point on the bottom-left position
+     * More details at:
+     * {@link FirebaseVisionText.TextBlock#getCornerPoints()},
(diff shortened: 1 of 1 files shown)


app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
```diff
@@ -0,0 +1,240 @@
...
+        FirebaseVisionTextRecognizer textRecognizer =
+                FirebaseVision.getInstance().getOnDeviceTextRecognizer();
+
+        /*
+        The standard java Semaphore is not used because it's not work properly
+        with the Task<FirebaseVisionText>.
+         */
+
+        // Instantiates the CountDownLatch used for synchronizing the extraction
+        final CountDownLatch extraction = new CountDownLatch(1); // Luca Moroldo (g3)
```
Giofasan: I know you did not write this line of code.
However I advise you not to use magic numbers, but create a variable [x = 1] and use this variable instead of the number [CountDownLatch (x)]

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
```diff
@@ -0,0 +1,240 @@
...
+     * Sorts the blocks recognized
+     * More details at: {@link FirebaseVisionText.TextBlock#getTextBlocks()}.
+     * @param OCRResult FirebaseVisionText object produced by an OCR recognition
+     * @return An ArrayList of FirebaseVisionText sorted
+     * @author Pietro Prandini (g2)
+     */
+    private ArrayList<FirebaseVisionText.TextBlock> sortBlocks(FirebaseVisionText OCRResult) {
+        ArrayList<FirebaseVisionText.TextBlock> OCRBlocks =
+                new ArrayList<>(OCRResult.getTextBlocks());
+        OCRBlocks = sortBlocksY(OCRBlocks);
```
Giofasan, line 187: I do not understand what sortBlocksY does

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
```diff
@@ -0,0 +1,240 @@
...
+    sortBlocksY(ArrayList<FirebaseVisionText.TextBlock> OCRBlocks) {
+        // Comparator for ordering the blocks by the y axis
+        Comparator<FirebaseVisionText.TextBlock> mYComparator =
+                new Comparator<FirebaseVisionText.TextBlock>() {
+            @Override
+            public int compare(FirebaseVisionText.TextBlock o1, FirebaseVisionText.TextBlock o2) {
+                int o1TopLeftY = Objects.requireNonNull(o1.getCornerPoints())[TOP_LEFT].y;
+                int o2TopLeftY = Objects.requireNonNull(o2.getCornerPoints())[TOP_LEFT].y;
+                return Integer.compare(o1TopLeftY,o2TopLeftY);
+            }
```
Giofasan, line 209: I do not understand why you compare only the blocks on the top left. Would not it be better to compare all blocks?

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
```diff
@@ -0,0 +1,240 @@
...
+        };
+        // Sorts the blocks
+        Collections.sort(OCRBlocks, mYComparator);
+        return OCRBlocks;
+    }
+}
```
Giofasan, line 215: Good comments and also well organized.
But I advise you not to exaggerate with the comments, as in the case of the variables TOP_LEFT, TOP_RIGHT etc. that you repeated the same comment for all four variables.

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
PrandiniUniPD: Thanks, I'll firstly put that number in a variable as you suggested.

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
PrandiniUniPD: To understand the functionality of that method we have to see the javadoc comments of that method, but I could write that javadoc better than this version.
Basically it prepares an ArrayList of the blocks of the text sorted from the block at the top to the block at the bottom.
Thanks for this comment, I'll add a comment before this statement and I'll write better the javadoc comments of the method sortBlocksY.

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
PrandiniUniPD: This is a comparator, I've put a link about that in the javadoc comment.
The real sorting process is at line 237 when the statement `Collections.sort(OCRBlocks, mYComparator)` puts in the ArrayList named OCRBlocks the blocks of the text and use the rule of mYComparator for sorting every block comparing every block with that rule.
I've chosen the TOP_LEFT.y of every block to compares the heights of the text in order to sort the blocks from top to bottom. Is it clear now? Could I make some comments to explain that better than now?
Thanks in advance.

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
PrandiniUniPD: Thanks for reviewing, I'll propose you another version with your suggestions soon.

PrandiniUniPD:
Now is it good for you?
Thanks in advance.

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
Giofasan: Yes thanks. Now with this explanation it is clearer

Giofasan:
Now it's much better than before. Good job

PR #44, opened 2018-11-30
CodeReview of 8 from 15

PietroPrandini:
From the code I've written, for this code review I've chosen to propose to you only the TextExtractor.java because it has a core functionality for the app.
There are many comments and links for supporting the review.
I hope you can review easily this code, if not, I'm here.
Have a good review.

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
@@ -1,0 +1,240 @@
+package unipd.se18.ocrcamera;
+
+import android.graphics.Bitmap;
+import android.util.Log;
+import com.google.android.gms.tasks.OnSuccessListener;
+import com.google.android.gms.tasks.Task;
+import com.google.firebase.ml.vision.FirebaseVision;
+import com.google.firebase.ml.vision.common.FirebaseVisionImage;
+import com.google.firebase.ml.vision.text.FirebaseVisionText;
+import com.google.firebase.ml.vision.text.FirebaseVisionTextRecognizer;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.CountDownLatch;
+
+/**
+ * Class that implements the common OCR interface for retrieving text from a Bitmap image.
+ * This class uses ml-kit provided by Firebase.
+ * More details at: {@link FirebaseVisionText}.
+ * @see <a href="https://firebase.google.com/docs/ml-kit/android/recognize-text">
+ *     ml-kit, recognize text</a>
+ * @author Pietro Prandini (g2)
+ */
+public class TextExtractor implements OCRInterface {
+    /**
+     * String used for the logs of this class.
+     */
+    private final String TAG = "TextExtractor";
+
+    /*
+    The next four int are used for recognizing the position of the FirebaseVisionText Objects.
+    They could be useful for sorting the blocks or for an automatic recognition
+    of the ingredients text block.
+     */
+
+    /**
+     * Index of the point on the top-left position.
+     * More details at:
+     * {@link FirebaseVisionText.TextBlock#getCornerPoints()},
+     * {@link FirebaseVisionText.Line#getCornerPoints()}
+     * or {@link FirebaseVisionText.Element#getCornerPoints()}.
+     */
+    private final int TOP_LEFT = 0;
+
+    /**
+     * Index of the point on the top-right position.
+     * More details at:
+     * {@link FirebaseVisionText.TextBlock#getCornerPoints()},
+     * {@link FirebaseVisionText.Line#getCornerPoints()}
+     * or {@link FirebaseVisionText.Element#getCornerPoints()}.
+     */
+    private final int TOP_RIGHT = 1;
+
+    /**
+     * Index of the point on the bottom-left position
+     * More details at:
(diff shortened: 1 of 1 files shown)


app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
```diff
@@ -0,0 +1,240 @@
...
+                                + "----- END OF THE RECOGNIZED TEXT -----");
+                        // Extraction ended - Analogous to signal
+                        extraction.countDown(); // Luca Moroldo (g3)
+                    }
+                });
+        // Waits until the extraction ends
+        if (!firebaseVisionTextTask.isSuccessful()) {
+            try {
+                // Analogous to wait
+                extraction.await(); // Luca Moroldo (g3)
```
stefanoromanello, line 135: With this you are waiting that the CountDownLatch reaches 0 and then the code can continue. The problem is that I don't see a way to identify if firebase gets an error.
The solution could be this:

```
Task<FirebaseVisionText> firebaseVisionTextTask = textRecognizer.processImage(firebaseVisionImage).addOnSuccessListener(new OnSuccessListener<FirebaseVisionText>() {
            @Override
            public void onSuccess(FirebaseVisionText firebaseVisionText) { Log.v(TAG, methodTag + "onSuccess ->\n" 
                    + "-----      RECOGNIZED TEXT       -----"
                    + "\n" + firebaseVisionText.getText() + "\n"
                    + "----- END OF THE RECOGNIZED TEXT -----");
                    // Extraction ended - Analogous to signal
                    extraction.countDown(); // Luca Moroldo (g3)
                }
        }).addOnFailureListener(new OnFailureListener() {
            @Override
            public void onFailure(@NonNull Exception e) {
                      //MANAGE FAILURE
                     extraction.countDown();
                }
        });
```

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
```diff
@@ -0,0 +1,240 @@
...
+     * Sorts the blocks recognized
+     * More details at: {@link FirebaseVisionText.TextBlock#getTextBlocks()}.
+     * @param OCRResult FirebaseVisionText object produced by an OCR recognition
+     * @return An ArrayList of FirebaseVisionText sorted
+     * @author Pietro Prandini (g2)
+     */
+    private ArrayList<FirebaseVisionText.TextBlock> sortBlocks(FirebaseVisionText OCRResult) {
+        ArrayList<FirebaseVisionText.TextBlock> OCRBlocks =
+                new ArrayList<>(OCRResult.getTextBlocks());
+        OCRBlocks = sortBlocksY(OCRBlocks);
```
stefanoromanello, line 192: Why did you split these 2 methods? For me they are not complicated to be splitted.

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
```diff
@@ -0,0 +1,240 @@
...
+    private final int TOP_LEFT = 0;
+
+    /**
+     * Index of the point on the top-right position.
+     * More details at:
+     * {@link FirebaseVisionText.TextBlock#getCornerPoints()},
+     * {@link FirebaseVisionText.Line#getCornerPoints()}
+     * or {@link FirebaseVisionText.Element#getCornerPoints()}.
+     */
+    private final int TOP_RIGHT = 1;
```
stefanoromanello: I suggest to delete these variables that are not used and relative comments for a better clean code

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
PrandiniUniPD: Thanks a lot, I'll add it!

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
PrandiniUniPD: I've split this method because the method for sorting could be change or basically in the near future could be useful sorting the blocks from left to right and not only from top to bottom.
So you suggest anyway to merge the methods or to left the methods split?
Thanks in advance.

stefanoromanello: Oh ok, no its fine. This is reasonable solution for the future

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
PrandiniUniPD: Thanks, I've left these variables for the near future about sorting blocks or recognizing the block of the ingredients. Could I remove eventually some comments?

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
PrandiniUniPD: Thanks for reviewing.

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
stefanoromanello: Yes, maybe you can group the comments into one like
> These variables are used for sorting the blocks. 

And if needed you can add more comments in the future methods

PrandiniUniPD: Ok, thanks, I'll push a commit with the requested changes soon.

PrandiniUniPD:
Now is it good?

PrandiniUniPD:
@stefanoromanello has said me that it's ok directly, so I'll close this pull request.

PR #45, opened 2018-11-30
codeReview of 2 from 9

mattia1142866:
the important part of code to review is in CameraActivity.java from line 202 to the end (in this moment the file is not ready to review, tomorrow morning (01/12) I review the code and it will ready to review).

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
@@ -299,15 +299,12 @@ public class CameraActivity extends AppCompatActivity {
         }else{
             Log.d("createOCRCDirs", "already exists");
         }
-        //Log.e("createTessDirs", "launching copyTessDataForTextRecognizor()");
-        //copyTessDataForTextRecognizor();
     }
 
     /**
      * Takes a photo.
      */
     private void takePhoto() {
-        Log.d(TAG,"takePicture");
         if(null == mCameraDevice) {
             Log.e(TAG, "cameraDevice is null");
             return;
@@ -319,12 +316,11 @@ public class CameraActivity extends AppCompatActivity {
             if (characteristics != null) {
                 jpegSizes = characteristics.get(CameraCharacteristics.SCALER_STREAM_CONFIGURATION_MAP).getOutputSizes(ImageFormat.JPEG);
             }
-            int width = 6440;
+            int width = 640;
             int height = 480;
             if (jpegSizes != null && 0 < jpegSizes.length) {
                 width = jpegSizes[0].getWidth();
                 height = jpegSizes[0].getHeight();
-                //devo fare un print delle dimensioni!!!!!!!!!!!!!!
             }
             ImageReader reader = ImageReader.newInstance(width, height, ImageFormat.JPEG, 1);
             List<Surface> outputSurfaces = new ArrayList<Surface>(2);
@@ -348,7 +344,7 @@ public class CameraActivity extends AppCompatActivity {
             } catch(IOException e){
                 e.printStackTrace();
             }
-            final Intent intentOCR= new Intent(this,ResultActivity.class);
+            final Intent intentOCR= new Intent(this, ResultActivity.class);
             intentOCR.putExtra("imgPath", file.getPath());
             ImageReader.OnImageAvailableListener readerListener = new ImageReader.OnImageAvailableListener() {
                 @Override
@@ -370,6 +366,12 @@ public class CameraActivity extends AppCompatActivity {
                         }
                     }
                 }
+
+                /**
+                 * saves the vector of bytes using FileOutputStream
+                 * @param bytes the vector of bytes of the bitmap image
+                 * @throws IOException because can occur a IOException
+                 */
                 private void save(byte[] bytes) throws IOException {
                     OutputStream output = null;
                     try {
@@ -390,7 +392,7 @@ public class CameraActivity extends AppCompatActivity {
                     Log.d(TAG, "foto salvata");
                     super.onCaptureCompleted(session, request, result);
                     Toast.makeText(CameraActivity.this, "Saved:" + file, Toast.LENGTH_SHORT).show();
-                    //createCameraPreview();
+                    //here was createCameraPreview() but in order to show the result i start another activity
                     startActivity(intentOCR);
                 }
             };
--- a/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
@@ -1,0 +1,446 @@
+package unipd.se18.ocrcamera;
+
+import android.Manifest;
+import android.content.Context;
+import android.content.Intent;
+import android.content.pm.PackageManager;
+import android.graphics.ImageFormat;
+import android.graphics.SurfaceTexture;
+import android.hardware.camera2.CameraAccessException;
+import android.hardware.camera2.CameraCaptureSession;
+import android.hardware.camera2.CameraCharacteristics;
+import android.hardware.camera2.CameraDevice;
+import android.hardware.camera2.CameraManager;
+import android.hardware.camera2.CameraMetadata;
+import android.hardware.camera2.CaptureRequest;
+import android.hardware.camera2.TotalCaptureResult;
+import android.hardware.camera2.params.StreamConfigurationMap;
+import android.media.Image;
+import android.media.ImageReader;
+import android.os.Bundle;
+import android.os.Environment;
+import android.os.Handler;
+import android.os.HandlerThread;
+import android.support.annotation.NonNull;
+import android.support.v4.app.ActivityCompat;
+import android.support.v7.app.AppCompatActivity;
+import android.util.Log;
+import android.util.Size;
+import android.util.SparseIntArray;
+import android.view.Surface;
+import android.view.TextureView;
+import android.view.View;
+import android.widget.Button;
+import android.widget.Toast;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Calendar;
+import java.util.List;
+
+/**
+ * The Activity useful for making photos
+ */
+public class CameraActivity extends AppCompatActivity {
+
+    // Final shared variables
+    /**
+     * TAG used for logs
+     */
+    private static final String TAG = "CameraActivity";
+    /**
+     * directory used to save images and/or text recognized
+     */
+    private static final File SAVE_DIR = new File(Environment.getExternalStorageDirectory().getAbsolutePath() + "/OCRCamera/Pictures/");
+
+    //creation intent
+
+    /**
+     * Code for the permission requested
+     */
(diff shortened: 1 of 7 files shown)


mattia1142866:
the last commit include more comments and clean code. this code is not working because i've not import the manifest and the resources.

leopardo-rossi-developer:
After the last commit you made I can't leave comments on your whole code because I can only read the last changes you made. For this reason to do the codereview I'll look the code you uploaded in the first commit, considering the updates you made in the second one

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,446 @@
...
+     * Controls the output of the permissions requests.
+     * @param requestCode The code assigned to the request
+     * @param permissions The list of the permissions requested
+     * @param grantResults The results of the requests
+     */
+    @Override
+    public void onRequestPermissionsResult(int requestCode, @NonNull String permissions[], @NonNull int[] grantResults) {
+        if (requestCode == MY_CAMERA_REQUEST_CODE) {
+            if (grantResults[0] == PackageManager.PERMISSION_DENIED) {
+                // close the app
```
leopardo-rossi-developer: Thinking to user's experience it's not good that if permissions aren't granted the app closes. 
Why don't you only show a message to the user to alert him that without permissions the app won't work?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,446 @@
...
+                Toast.makeText(CameraActivity.this, "Sorry!!!, you can't use this app without granting permission", Toast.LENGTH_LONG).show();
+                finish();
```
leopardo-rossi-developer, line 237: Thinking to user's experience it's not good that if permissions aren't granted the app closes.
Why don't you only show a message to the user to alert him that without permissions the app won't work?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,446 @@
...
+
+    /**
+     * CameraPatureSession used for the camera preview;
+     */
+    private CameraCaptureSession mCameraCaptureSession;
+
+    /**
+     * Size used for the camera preview
+     */
+    private Size mPreviewSize;
```
leopardo-rossi-developer, line 109: I think this name isn't clear enough. I would have used something like "cameraPreviewSize"

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,446 @@
...
+        try {
+            //creates texture view
+            SurfaceTexture texture = mCameraTextureView.getSurfaceTexture();
+            assert texture != null;
+            texture.setDefaultBufferSize(mPreviewSize.getWidth(), mPreviewSize.getHeight());
+            //Output
+            Surface surface = new Surface(texture);
+            mCaptureRequestBuilder = mCameraDevice.createCaptureRequest(CameraDevice.TEMPLATE_PREVIEW);
+            mCaptureRequestBuilder.addTarget(surface);
+            mCameraDevice.createCaptureSession(Arrays.asList(surface), new CameraCaptureSession.StateCallback(){
```
leopardo-rossi-developer, line 255: Here comments aren't clear enough. 
Which is _texture_ variable purpose? Which is _surface_ variable purpose?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,446 @@
...
+    /**
+     * Updates the preview of the camera frames.
+     */
+    private void updatePreview() {
+        if(null == mCameraDevice) {
+            Log.e(TAG, "updatePreview error, return");
+        }
+        mCaptureRequestBuilder.set(CaptureRequest.CONTROL_MODE, CameraMetadata.CONTROL_MODE_AUTO);
+        try {
+            mCameraCaptureSession.setRepeatingRequest(mCaptureRequestBuilder.build(), null, mBackgroundHandler);
```
leopardo-rossi-developer, line 285: What are you doing in this line and in line 283? There's no comments in this part of code and I think you should insert them because this is an important part of code

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,446 @@
...
+        } catch (CameraAccessException e) {
+            e.printStackTrace();
+        }
+    }
+
+    /**
+     * Create App's Directories
+     */
+    public void createOCRCDirs(){
```
leopardo-rossi-developer, line 294: For what reason do you want to create directories for OCR? If I were you, I'd add more accurate method's specification so that it's clear for who read this method which is its purpose?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,446 @@
...
+        }
+        CameraManager manager = (CameraManager) getSystemService(Context.CAMERA_SERVICE);
+        try {
+            CameraCharacteristics characteristics = manager.getCameraCharacteristics(mCameraDevice.getId());
+            Size[] jpegSizes = null;
+            if (characteristics != null) {
+                jpegSizes = characteristics.get(CameraCharacteristics.SCALER_STREAM_CONFIGURATION_MAP).getOutputSizes(ImageFormat.JPEG);
+            }
+            int width = 6440;
+            int height = 480;
```
leopardo-rossi-developer, line 320: I'll suggest you to put these two numbers into two constants, because in this reason if in the future they will change it will be easier to update them

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,446 @@
...
+            if (jpegSizes != null && 0 < jpegSizes.length) {
```
leopardo-rossi-developer, line 321: Ambiguous if condition. What do you think of _jpegSizes != null && jpegSizes.length > 0_?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,446 @@
...
+                height = jpegSizes[0].getHeight();
+                //devo fare un print delle dimensioni!!!!!!!!!!!!!!
+            }
+            ImageReader reader = ImageReader.newInstance(width, height, ImageFormat.JPEG, 1);
+            List<Surface> outputSurfaces = new ArrayList<Surface>(2);
+            outputSurfaces.add(reader.getSurface());
+            outputSurfaces.add(new Surface(mCameraTextureView.getSurfaceTexture()));
+            final CaptureRequest.Builder captureBuilder = mCameraDevice.createCaptureRequest(CameraDevice.TEMPLATE_STILL_CAPTURE);
+            captureBuilder.addTarget(reader.getSurface());
+            captureBuilder.set(CaptureRequest.CONTROL_MODE, CameraMetadata.CONTROL_MODE_AUTO);
```
leopardo-rossi-developer, line 331: This part of code, from line 329 to line 335, is very hard to understand without any comments

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,446 @@
...
+            final File file = new File(Environment.getExternalStorageDirectory()+"/OCRCamera/Pictures/"+currentTime+".jpg");
+            file.mkdirs();
+            try {
+                if(file.exists()){
+                    file.delete();
+                }
+                file.createNewFile();
+            } catch(IOException e){
+                e.printStackTrace();
+            }
```
leopardo-rossi-developer, line 346: Why don't you put this part of code, from line 341 to line 350, inside a method that will return the file's path? In this way the code will be clearer than now and you can use file's creation in different parts of code without writing it every time.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,446 @@
...
+            };
+            reader.setOnImageAvailableListener(readerListener, mBackgroundHandler);
+            final CameraCaptureSession.CaptureCallback captureListener = new CameraCaptureSession.CaptureCallback() {
+                @Override
+                public void onCaptureCompleted(CameraCaptureSession session, CaptureRequest request, TotalCaptureResult result) {
+                    Log.d(TAG, "foto salvata");
+                    super.onCaptureCompleted(session, request, result);
+                    Toast.makeText(CameraActivity.this, "Saved:" + file, Toast.LENGTH_SHORT).show();
+                    //createCameraPreview();
+                    startActivity(intentOCR);
```
leopardo-rossi-developer, line 396: If I were you, I'd declared the intente variable here and not where you've declared it because you need it at this point of code and not before. In this way another programmer that reads this code doesn't have to scroll up trying to find where _intentOCR_ is declared

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,446 @@
...
+    }
+    protected void stopBackgroundThread() {
+        mBackgroundThread.quitSafely();
+        try {
+            mBackgroundThread.join();
+            mBackgroundThread = null;
+            mBackgroundHandler = null;
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
```
leopardo-rossi-developer, line 446: These two methods don't have any description to say to who read them what is their purpose

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
mattia1142866: right observation. maybe i tougth that was the fastest way to write the code, and after i've never optimized it

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
mattia1142866: you're right. my mistake.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
mattia1142866: beacouse my goal was to keeping track keep track of many processed photos (at least the under 50)

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
mattia1142866: possibile (but really difficult) computational errors.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
mattia1142866: yes, you're right. in this class it's quiet ambiguous, i thought that was useful to find the errors and what triggered them, in this case concerning the creation of directories.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
leopardo-rossi-developer: Ok perfect! Write this thing in method's specifications so that is it clear for everyone who reads the method

leopardo-rossi-developer:
I can't close the code review because I can't see a commit with the changes I suggested you

PR #46, opened 2018-11-30
CodeReview of 18 from 4

Fisher4537:
Implemented Barcode Scanner with [MLKit](https://firebase.google.com/docs/ml-kit/android/read-barcodes) library.
Hello Porro, this is a copy of the pull request that you can find [here](https://github.com/PrandiniUniPD/OCRCamera/pull/34). For the code review it is y not necessary that you review all the changes: it would be too much. You can ignore all the .xml file changes and review ONLY the .java files. Let me know if you have better proposal.


app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -33,6 +44,8 @@
      */
     private TextView mOCRTextView;
 
+    private TextView mBarcodeView;
```
thomasporro: Maybe here you can write what's mBarcodeView is like it's done in line 43 for mOCRTextView, and in line 50 for last photo.

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -56,16 +73,18 @@ public void onClick(View view) {
             }
         });
 
-
         //Get image path and text of the last image from preferences
         SharedPreferences prefs = getSharedPreferences("prefs", MODE_PRIVATE);
         String pathImage = prefs.getString("imagePath", null);
         String OCRText = prefs.getString("text", null);
+        String barcodeText = prefs.getString("barcodeText",null);
```
thomasporro, line 83: Maybe you can modify line 76 to add the extracting the last barcode read in the comment

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -180,5 +213,79 @@ protected void onPreExecute() {
...
+
+        @Override
+        protected void onPreExecute() {
+            progressDialog = ProgressDialog.show(ResultActivity.this, progressMessage, "Barcode detection...");
+        }
+
+        @Override
+        protected String doInBackground(Bitmap... bitmaps){
+            String rawBarcode = "";
+            //BarcodeExtractor barcodeExtractor = ;
```
thomasporro: It's just my opinion, but here maybe you can specify why the method is commented. So it's easier to understand if you want to modify the code in future or just you deleted the line

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -180,5 +213,79 @@ protected void onPreExecute() {
...
+     * @author Luca Perali (g4) (inspired to AsyncLoad class)
+     */
+    private class BarcodeReader extends AsyncTask<Bitmap, Void, String>{
+
+        protected ProgressDialog progressDialog; //TODO: progressDialog is deprecated, replace with ProgressBar
+        protected TextView barcodeTextView; //
+        protected String progressMessage;
+
+        /**
+         * Costructor initialize:
```
thomasporro: Maybe you can specify which variables the constructor initialize.

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -180,5 +213,79 @@ protected void onPreExecute() {
...
+    /**
+     * Read the barcode in background and post the results on barcodeView.
+     * Documentation of AsyncTask at https://developer.android.com/reference/android/os/AsyncTask
+     *
+     * @author Luca Perali (g4) (inspired to AsyncLoad class)
+     */
+    private class BarcodeReader extends AsyncTask<Bitmap, Void, String>{
+
+        protected ProgressDialog progressDialog; //TODO: progressDialog is deprecated, replace with ProgressBar
+        protected TextView barcodeTextView; //
```
thomasporro: You probably forgot to delete the "//"

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -180,5 +213,79 @@ protected void onPreExecute() {
...
+
+        @Override
+        protected String doInBackground(Bitmap... bitmaps){
+            String rawBarcode = "";
+            //BarcodeExtractor barcodeExtractor = ;
+
+            if(lastPhoto != null) {
+                rawBarcode = new BarcodeExtractor().getTextFromImg(lastPhoto); //see BarcodeExtractor class
+
+                if(rawBarcode.equals("")) //No barcode found case
```
thomasporro: In my opinion the brace should be put in the same line of the "if". Here a my suggestion
```suggestion
                //No barcode found case
                if(rawBarcode.equals("")){
```

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -180,5 +213,79 @@ protected void onPreExecute() {
...
+                {
+                    final String finalTextRecognized = getString(R.string.barcode_not_found);
+                    runOnUiThread(new Runnable() { //asynchronous thread to optimize performance
+                        @Override
+                        public void run() {
+                            barcodeTextView.setText(finalTextRecognized);
+                        }
+                    });
+                }
```
thomasporro: The same of the above comment. I leave down below a suggestion:
```suggestion
                } else {    // found at least one barcode case
```

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
```diff
@@ -25,7 +25,9 @@
     /**
      * Constructor
      */
-    TextExtractor() {    }
+    TextExtractor() {
```
thomasporro, line 28: It's just a my preference, but i prefer the constructor and the braces in the same line

app/src/main/java/unipd/se18/ocrcamera/TextExtractor.java
```diff
@@ -54,23 +56,24 @@ private String extractText(Bitmap img)
...
+                                + firebaseVisionText.getText() + "\n----- END OF THE RECOGNIZED TEXT -----");
 
-                //analogous to signal
-                latch.countDown();
-            }
-        });
+                        //analogous to signal
+                        latch.countDown();
+                    }
+                });
```
thomasporro, line 76: The same question I've asked in BarcodeExtractor.java

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
```diff
@@ -0,0 +1,119 @@
...
+import com.google.firebase.ml.vision.barcode.FirebaseVisionBarcode;
+import com.google.firebase.ml.vision.barcode.FirebaseVisionBarcodeDetector;
+import com.google.firebase.ml.vision.common.FirebaseVisionImage;
+
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+
+/**
+ * Class which try to extract the content of a barcode in an image
+ * @author Luca Perali (inspired to Pietro Prandini's TextExtractor)
```
thomasporro, line 16: Nice referral

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
```diff
@@ -0,0 +1,119 @@
...
+        //Nothing to initialize yet
+    }
+
+    /**
+     * This method try to extract barcode information from an image if can,
+     * otherwise return an empty string.
+     * (The method has the same name of OCRInterface methods for eventual future implementation of a common
+     * interface)
+     *
+     * @param img Bitmap image with or without barcode in it
```
thomasporro, line 38: Maybe you can modify the specs, because "image with or without barcode in it", is just the image we want to analize. The fact that if there is a barcode is already reported in @return

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
```diff
@@ -0,0 +1,119 @@
...
+     * @return string with barcode info if barcode is found, empty if no barcode is found
+     * @author Luca Perali
+     */
+    public String getTextFromImg(Bitmap img){
+
+        Log.d(TAG, "extractTextFromImage");
+        long initialTime = java.lang.System.currentTimeMillis();
+
+        // Firebase implementation tutorial at https://firebase.google.com/docs/ml-kit/android/read-barcodes
```
thomasporro, line 47: The links to the documentations are very nice

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
```diff
@@ -0,0 +1,119 @@
...
+                        Log.v(TAG, "extractText -> onSuccess ->\n" +
+                                "-----      RECOGNIZED BARCODE       -----\n"
+                                + getInfoFromBarcode(barcodes) + "\n" +
+                                "----- END OF THE RECOGNIZED TEXT -----");
+
+                        //analogous to signal
+                        latch.countDown();
+                    }
+                });
+        /*         .addOnFailureListener(new OnFailureListener() {
```
thomasporro: It's just my opinion, but here maybe you can specify why the method is commented. So it's easier to understand if you want to modify the code in future or just you deleted the line

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
```diff
@@ -0,0 +1,119 @@
...
+                    public void onSuccess(List<FirebaseVisionBarcode> barcodes) {
+                        Log.v(TAG, "extractText -> onSuccess ->\n" +
+                                "-----      RECOGNIZED BARCODE       -----\n"
+                                + getInfoFromBarcode(barcodes) + "\n" +
+                                "----- END OF THE RECOGNIZED TEXT -----");
+
+                        //analogous to signal
+                        latch.countDown();
+                    }
+                });
```
thomasporro, line 70: This is a question: the brace reffered to .addOnSuccessListener or Task<List<FirebaseVisionBarcode>>?

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
```diff
@@ -0,0 +1,119 @@
...
+                        @Override
+                        public void onFailure(@NonNull Exception e) {
+                            // Task failed with an exception
+                            // ...
+                        }
+                    });*/
+
+        // if barcode is not detected yet
+        if(!task.isSuccessful()) {
+            try
```
thomasporro: You should use the syntax you've already used in other parts of the code. Like this:

```suggestion
           try{ 
                //the things you want to do
           } else {
                //other things
           }
```

thomasporro: And the same is for the catch statement

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Get a string from the Firebase barcodes list object.
+     *
+     * @param barcodes rapresents the list of barcode detected by Firebase library
+     * @return the string conversion of the input param, empty string if no barcode is detected
+     * @author Luca Perali
+     */
+    private static String getInfoFromBarcode(List<FirebaseVisionBarcode> barcodes) {
+        //StringBuilder result = new StringBuilder();
```
thomasporro: It's just my opinion, but here maybe you can specify why these methods are commented. So it's easier to understand if you want to modify the code in future, you used for debugging or just you deleted the lines

thomasporro (commented):
Sorry if I took so long to do the review but in these days I was very busy.
Btw the code is ok, I've commented some parts where the syntax is different from the code in general, and some parts that contains methods-comments.
But in general the code is very good.

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
Fisher4537: What do you think about "barcodeTextView"? It seems better to me!

thomasporro: I was thinking to simply add a comment above your variable to make it more readable

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
Fisher4537: Thanks, I've delated the comment! It has been my forgetfulness

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
Fisher4537: Which comment are you talking about? Do you mean that barcodeText should be the string of the last barcode readed?

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
Fisher4537: Thanks! I think that it's very usefull to have a pointer on the same documentation!

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
Fisher4537: You are right! I'm Fixing it! Thanks

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
thomasporro: The comment in line 76. Just like this:
`\\Get the image path, text and barcode of the last image from preferences `

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
Fisher4537: image
Well, as you see in the image the yellow braces contain the explicit parameters of addOnSuccessListener method

thomasporro: 👍🏼

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
Fisher4537: Bad abit! Thanks for the correction. Added for the next commit!

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
Fisher4537: I agree! I forgot also this one commented line! Thanks

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
Fisher4537: Oh nice! Now I can see what you mean!

Fisher4537:
Thank to you, I've fixed the document style bug and deleted useless comments. I'm going to commit the last comment modification and then if you agree we could close the pull request!

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
```diff
@@ -0,0 +1,118 @@
+package unipd.se18.ocrcamera;
+
+
```
thomasporro: They reported in my review that according to our style document:

>Imports are ordered as follows:
>All static imports in a single block.
>All non-static imports in a single block.
>If there are both static and non-static imports, a single blank line separates the two blocks. There are no other blank lines between import statements.

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -33,6 +44,8 @@
      */
     private TextView mOCRTextView;
```
thomasporro, line 46: I was thinking to add an "explanation" like this:
```suggestion
     /**
      * The TextView of the extracted barcode from the captured photo.
      */
```

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
thomasporro: The comment is still there

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
thomasporro: I would prefer if the commented methods had an explanation, for example if you made them for debugging

thomasporro (commented):
The code seems better, there are only a few things to change and in my opinion it's ok

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
Fisher4537: Sure! Thanks!

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
Fisher4537: I've changed the name in "barcodeTextView" and added a comment, to the variable to make it clearer!

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
Fisher4537: Thanks for the suggestion! I'm fixing it!

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
thomasporro: This comment  ( `//StringBuilder result = new StringBuilder();` ) and the following in my opinion still a problem in the code until you don't explain their function or you delete them

thomasporro (commented):
The changes you made were fine! There is only a few commented methods to fix

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
Fisher4537: Pushing the remotion of the comments!

Fisher4537:
Thanks for the review Thomas, I appreciated your tenacity! Closing the pull request

PR #41, opened 2018-11-30
CodeReview of 9 from 16

leopardo-rossi-developer:
The method's for the code review are: takePhoto() and tempFileImage().

PR #47, opened 2018-11-30
CodeReview of 2 from 15

mattia1142866:
the important part of code to review is in CameraActivity.java from line 202 to the end (in this moment the file is not ready to review, tomorrow morning (01/12) I review the code and it will ready to review).

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
@@ -299,15 +299,12 @@ public class CameraActivity extends AppCompatActivity {
         }else{
             Log.d("createOCRCDirs", "already exists");
         }
-        //Log.e("createTessDirs", "launching copyTessDataForTextRecognizor()");
-        //copyTessDataForTextRecognizor();
     }
 
     /**
      * Takes a photo.
      */
     private void takePhoto() {
-        Log.d(TAG,"takePicture");
         if(null == mCameraDevice) {
             Log.e(TAG, "cameraDevice is null");
             return;
@@ -319,12 +316,11 @@ public class CameraActivity extends AppCompatActivity {
             if (characteristics != null) {
                 jpegSizes = characteristics.get(CameraCharacteristics.SCALER_STREAM_CONFIGURATION_MAP).getOutputSizes(ImageFormat.JPEG);
             }
-            int width = 6440;
+            int width = 640;
             int height = 480;
             if (jpegSizes != null && 0 < jpegSizes.length) {
                 width = jpegSizes[0].getWidth();
                 height = jpegSizes[0].getHeight();
-                //devo fare un print delle dimensioni!!!!!!!!!!!!!!
             }
             ImageReader reader = ImageReader.newInstance(width, height, ImageFormat.JPEG, 1);
             List<Surface> outputSurfaces = new ArrayList<Surface>(2);
@@ -348,7 +344,7 @@ public class CameraActivity extends AppCompatActivity {
             } catch(IOException e){
                 e.printStackTrace();
             }
-            final Intent intentOCR= new Intent(this,ResultActivity.class);
+            final Intent intentOCR= new Intent(this, ResultActivity.class);
             intentOCR.putExtra("imgPath", file.getPath());
             ImageReader.OnImageAvailableListener readerListener = new ImageReader.OnImageAvailableListener() {
                 @Override
@@ -370,6 +366,12 @@ public class CameraActivity extends AppCompatActivity {
                         }
                     }
                 }
+
+                /**
+                 * saves the vector of bytes using FileOutputStream
+                 * @param bytes the vector of bytes of the bitmap image
+                 * @throws IOException because can occur a IOException
+                 */
                 private void save(byte[] bytes) throws IOException {
                     OutputStream output = null;
                     try {
@@ -390,7 +392,7 @@ public class CameraActivity extends AppCompatActivity {
                     Log.d(TAG, "foto salvata");
                     super.onCaptureCompleted(session, request, result);
                     Toast.makeText(CameraActivity.this, "Saved:" + file, Toast.LENGTH_SHORT).show();
-                    //createCameraPreview();
+                    //here was createCameraPreview() but in order to show the result i start another activity
                     startActivity(intentOCR);
                 }
             };
--- a/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
@@ -1,0 +1,446 @@
+package unipd.se18.ocrcamera;
+
+import android.Manifest;
+import android.content.Context;
+import android.content.Intent;
+import android.content.pm.PackageManager;
+import android.graphics.ImageFormat;
+import android.graphics.SurfaceTexture;
+import android.hardware.camera2.CameraAccessException;
+import android.hardware.camera2.CameraCaptureSession;
+import android.hardware.camera2.CameraCharacteristics;
+import android.hardware.camera2.CameraDevice;
+import android.hardware.camera2.CameraManager;
+import android.hardware.camera2.CameraMetadata;
+import android.hardware.camera2.CaptureRequest;
+import android.hardware.camera2.TotalCaptureResult;
+import android.hardware.camera2.params.StreamConfigurationMap;
+import android.media.Image;
+import android.media.ImageReader;
+import android.os.Bundle;
+import android.os.Environment;
+import android.os.Handler;
+import android.os.HandlerThread;
+import android.support.annotation.NonNull;
+import android.support.v4.app.ActivityCompat;
+import android.support.v7.app.AppCompatActivity;
+import android.util.Log;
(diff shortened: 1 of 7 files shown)


stefanoromanello:
Have you also done the takePhoto() method? Conflicts in #42 #41

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,448 @@
...
+        CameraManager manager = (CameraManager) getSystemService(Context.CAMERA_SERVICE);
+        Log.e(TAG, "is camera open");
+        try {
+            mCameraId = manager.getCameraIdList()[0];
+            CameraCharacteristics characteristics = manager.getCameraCharacteristics(mCameraId);
+            StreamConfigurationMap map = characteristics.get(CameraCharacteristics.SCALER_STREAM_CONFIGURATION_MAP);
+            assert map != null;
+            mPreviewSize = map.getOutputSizes(SurfaceTexture.class)[0];
+            // Add permission for camera and let user grant the permission
+            if (ActivityCompat.checkSelfPermission(this, Manifest.permission.CAMERA) != PackageManager.PERMISSION_GRANTED && ActivityCompat.checkSelfPermission(this, Manifest.permission.WRITE_EXTERNAL_STORAGE) != PackageManager.PERMISSION_GRANTED) {
```
stefanoromanello, line 215: I suggest to write this part as the first thing of this method. I don't see the need to execute the previous lines if I don't have permissions.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,448 @@
...
+                ActivityCompat.requestPermissions(CameraActivity.this, new String[]{Manifest.permission.CAMERA, Manifest.permission.WRITE_EXTERNAL_STORAGE}, MY_CAMERA_REQUEST_CODE);
+                return;
```
stefanoromanello, line 217: Shouldn't be better to write an alert or some message to advise the user before the return?
like onRequestPermissionsResult()

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,448 @@
...
+            }, null);
+        } catch (CameraAccessException e) {
+            e.printStackTrace();
+        }
+    }
+
+    /**
+     * Updates the preview of the camera frames.
+     */
+    private void updatePreview() {
```
stefanoromanello, line 279: Why this part is not in createCameraPreview() where is called? There are no parameters either returns of value so I don't see the need to take it out from the previous method. Maybe is for other reasons that I don't see

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,448 @@
...
+
+    // Final shared variables
+    /**
+     * TAG used for logs
+     */
+    private static final String TAG = "CameraActivity";
+    /**
+     * directory used to save images and/or text recognized
+     */
+    private static final File SAVE_DIR = new File(Environment.getExternalStorageDirectory().getAbsolutePath() + "/OCRCamera/Pictures/");
```
stefanoromanello, line 60: This would be more useful if its a string value because you can extend (removing final) the path without working with the file object.
Also is there a need for the static attribute?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,448 @@
...
+        } catch (CameraAccessException e) {
+            e.printStackTrace();
+        }
+    }
+
+    /**
+     * Create App's Directories
+     */
+    public void createOCRCDirs(){
+        File f = new File(Environment.getExternalStorageDirectory().getAbsolutePath()+"/OCRCamera/Pictures");
```
stefanoromanello, line 295: I would have used the global variable SAVE_DIR. Is there a reason to keep as it is?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,448 @@
...
+            outputSurfaces.add(new Surface(mCameraTextureView.getSurfaceTexture()));
+            final CaptureRequest.Builder captureBuilder = mCameraDevice.createCaptureRequest(CameraDevice.TEMPLATE_STILL_CAPTURE);
+            captureBuilder.addTarget(reader.getSurface());
+            captureBuilder.set(CaptureRequest.CONTROL_MODE, CameraMetadata.CONTROL_MODE_AUTO);
+            // Orientation
+            int rotation = getWindowManager().getDefaultDisplay().getRotation();
+            captureBuilder.set(CaptureRequest.JPEG_ORIENTATION, ORIENTATIONS.get(rotation));
+            // file name
+            int currentTime = Calendar.getInstance().getTime().hashCode();
+            final File file = new File(Environment.getExternalStorageDirectory()+"/OCRCamera/Pictures/"+currentTime+".jpg");
```
stefanoromanello, line 337: I would have used the global variable SAVE_DIR and extend it to point it to the photo. Is there a reason to keep as it is?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,448 @@
...
+            return;
+        }
+        CameraManager manager = (CameraManager) getSystemService(Context.CAMERA_SERVICE);
+        try {
+            CameraCharacteristics characteristics = manager.getCameraCharacteristics(mCameraDevice.getId());
+            Size[] jpegSizes = null;
+            if (characteristics != null) {
+                jpegSizes = characteristics.get(CameraCharacteristics.SCALER_STREAM_CONFIGURATION_MAP).getOutputSizes(ImageFormat.JPEG);
+            }
+            int width = 640;
```
stefanoromanello, line 319: I see this as problem because is hardcore programming. I suggest to find a way to get the resolution of the camera if is possible and use that variable as width and height. What do you think?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,448 @@
...
+     * Creates the preview of the camera capture frames.
+     */
+    private void createCameraPreview() {
+        try {
+            //creates texture view
+            SurfaceTexture texture = mCameraTextureView.getSurfaceTexture();
+            assert texture != null;
+            texture.setDefaultBufferSize(mPreviewSize.getWidth(), mPreviewSize.getHeight());
+            //Output
+            Surface surface = new Surface(texture);
```
stefanoromanello, line 252: This part is difficult to understand (few comments). What is the object mCameraTextureView? because you have created a surface object passing a SurfaceTexture parameter which used mCameraTextureView.getSurfaceTexture(). Is it possible to ha a clarification please?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
mattia1142866: i thought that the previous lines don't need the permissions. also you have to consider that the permission are controlled before, this control is for safety

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
mattia1142866: yes, this is the directory "root" for the application

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
mattia1142866: no, excuse me. when i tried to resolve a problems occured i changed this line and after i've never changed again

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
mattia1142866: the same answer before

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
mattia1142866: i thought of that, and reading on internet i read that is better to save images with the preffered sizes of the format and not of the camera. but it is a good advice in order of preprocessing images

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,448 @@
...
+            Log.d("createOCRCDirs", "CREATED");
+        }else{
+            Log.d("createOCRCDirs", "already exists");
+        }
+    }
+
+    /**
+     * Takes a photo.
+     */
+    private void takePhoto() {
```
stefanoromanello, line 307: TakePhoto is a long method, I suggest to split in more parts for a better testing.

stefanoromanello:
Last thing, is also missing a great part of specifications for the methods.  Have you the specificiation into another branch?

mattia1142866:
I will take into consideration the comments received

Issue #37, opened 2018-11-29
Photo_button not working on emulator.


leopra:
On line 142 of CameraActivity there is a problem because the code checks orientationResult but on an emulator it is always setted as NULL as this parameter is setted with the onSensorChange method. In this way it never enters the if condition and the ResultActivity never starts.
For now i have disabled that IF to continue working on the application.
Credits: dottor Li

frankplus:
I merged a fix, can you please verify that it is now working on the emulator?

PR #50, opened 2018-11-30
Avoid null value to orientationResult

frankplus:
This simple change may solve the problem stated in issue #37


app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -26,7 +26,7 @@
 public class CameraActivity extends AppCompatActivity {
 
     private CameraKitView cameraKitView;
-    private static String orientationResult;
+    private static String orientationResult="P";
```
PrandiniUniPD: It seems this change make sense!

PrandiniUniPD:
It seems this change make sense!

thomasporro (approved):
I agree with @PrandiniUniPD

PR #51, opened 2018-11-30
CodeReview to 18 from 10

Fisher4537:
Implemented Barcode Scanner with MLKit library.
Hello Alberto, this is a copy of the pull request that you can find here. For the code review it is not necessary that you review all the changes: it would be too much. You can ignore all the .xml file changes and review ONLY the .java files. Let me know if you have better proposal.


app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
```diff
@@ -0,0 +1,119 @@
...
+     */
+    public String getTextFromImg(Bitmap img){
+
+        Log.d(TAG, "extractTextFromImage");
+        long initialTime = java.lang.System.currentTimeMillis();
+
+        // Firebase implementation tutorial at https://firebase.google.com/docs/ml-kit/android/read-barcodes
+        FirebaseVisionImage image = FirebaseVisionImage.fromBitmap(img);
+        FirebaseVisionBarcodeDetector detector = FirebaseVision.getInstance()
+                .getVisionBarcodeDetector();
```
AlbVal97: Nice line-wrapping!

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
```diff
@@ -0,0 +1,119 @@
...
+
+        //latch used to wait barcode extraction to complete
+        //https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CountDownLatch.html
+        final CountDownLatch latch = new CountDownLatch(1);
+
+        Task<List<FirebaseVisionBarcode>> task = detector.detectInImage(image) // this start an asynchronous thread
```
AlbVal97: This line exceeds the column limit according to our Style document. May it be a simple solution to move the comment over the task declaration?

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
```diff
@@ -0,0 +1,119 @@
...
+                .addOnSuccessListener(new OnSuccessListener<List<FirebaseVisionBarcode>>() {
+                    @Override
+                    public void onSuccess(List<FirebaseVisionBarcode> barcodes) {
+                        Log.v(TAG, "extractText -> onSuccess ->\n" +
+                                "-----      RECOGNIZED BARCODE       -----\n"
+                                + getInfoFromBarcode(barcodes) + "\n" +
+                                "----- END OF THE RECOGNIZED TEXT -----");
+
+                        //analogous to signal
```
AlbVal97, line 69: The comment was meant to say that, according to CountDownLatch documentation, the method countDown() down here works as a "signal()" method applied to a semaphore, wasn't it? May it result ambiguous or hard to understand?

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
```diff
@@ -0,0 +1,119 @@
...
+        /*         .addOnFailureListener(new OnFailureListener() {
+                        @Override
+                        public void onFailure(@NonNull Exception e) {
+                            // Task failed with an exception
+                            // ...
+                        }
+                    });*/
+
+        // if barcode is not detected yet
+        if(!task.isSuccessful()) {
```
AlbVal97, line 82: I don't understand why the following try-catch statement doesn't follow braces rules according to our Style document. Was there a proper distinctive reason I haven't seen?

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -56,16 +73,18 @@ public void onClick(View view) {
...
         String OCRText = prefs.getString("text", null);
+        String barcodeText = prefs.getString("barcodeText",null);
 
         lastPhoto = BitmapFactory.decodeFile(pathImage);
 
         if (lastPhoto != null) {
-            mImageView.setImageBitmap(Bitmap.createScaledBitmap(lastPhoto, lastPhoto.getWidth(), lastPhoto.getHeight(), false));
+            mImageView.setImageBitmap(Bitmap.createScaledBitmap(lastPhoto, lastPhoto.getWidth(),
+                            lastPhoto.getHeight(),
+                            false));
```
AlbVal97, line 87: Nice line-wrapping!

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -180,5 +213,79 @@ protected void onPreExecute() {
...
+     */
+    private class BarcodeReader extends AsyncTask<Bitmap, Void, String>{
+
+        protected ProgressDialog progressDialog; //TODO: progressDialog is deprecated, replace with ProgressBar
+        protected TextView barcodeTextView; //
+        protected String progressMessage;
+
+        /**
+         * Costructor initialize:
+         * @param barcodeTextView the View responsable for barcode rapresentation
```
AlbVal97: I'm not sure about "responsable" and "rapresentation". May it be that some vowels have been scrambled?

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -180,5 +213,79 @@ protected void onPreExecute() {
...
+        }
+
+        @Override
+        protected String doInBackground(Bitmap... bitmaps){
+            String rawBarcode = "";
+            //BarcodeExtractor barcodeExtractor = ;
+
+            if(lastPhoto != null) {
+                rawBarcode = new BarcodeExtractor().getTextFromImg(lastPhoto); //see BarcodeExtractor class
+
```
AlbVal97, line 253: I don't understand why the following if-else statement doesn't follow braces rules according to our Style document.

AlbVal97 (commented):
The code is well built and easy to read: great work! The only inaccuracies I may have found are about strict following of our Style document rules. Furthermore, a lot of comments contain little "bad English" mistakes I have chosen not to point out because rarely they compromise code readablity, but I equally suggest to edit to improve the overall quality.

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
Fisher4537: Since the link to the CountDownLatch documentation has been passed, would it be clearer by omitting "//analogous to signal" comment?
I think also "//decrement CountDownLatch by one" would be ok! What do you think about it?

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
Fisher4537: I didn't pay attantion to the style because the proposed style document isn't approved yet! However the style is different from the rest of the code! I'm going to fix it! Thanks!

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
Fisher4537: Do you mean that the words aren't correct? What do you suggest?

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
Fisher4537: I didn't pay attantion to the style because the proposed style document isn't approved yet! However the style is different from the rest of the code! I'm going to fix it! Thanks!

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
AlbVal97: Yes, I think your last proposal should just be fine and less ambiguous. This way the reader could decide whether to read the documentation or not at his own risk

app/src/main/java/unipd/se18/ocrcamera/BarcodeExtractor.java
AlbVal97: Ok, so for the rest of the review I followed the not-yet-approved Style document down here:

[Documento di stile.docx](https://github.com/PrandiniUniPD/OCRCamera/files/2637127/Documento.di.stile.docx)

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
AlbVal97: Exactly, I would replace "responsable" with "responsible" and "rapresentation" with "representation"

Fisher4537:
Thank to you, I've fixed the document style bug and deleted useless comments. I'm going to commit the last comment modification where I remove also some hardcode and then if you agree we could close the pull request!

AlbVal97 (approved):
Ok, I think all has been solved

PR #53, opened 2018-12-01
CodeReview of 14 from 0

frankplus:
Hello Fasan,
here is some of my code that needs to be reviewed.
could you please review:
in PhotoTester class
1) float ingredientsTextComparison(String correct, String extracted); 
2) String getTagsStatsString()
3) String getMinKey(Map<String, Float> map)

and TestElement Class

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
@@ -429,6 +429,11 @@ public class PhotoTester {
         return report;
     }
 
+    /**
+     * Find and return key corresponding to minimum value
+     * @param map
+     * @return key corresponding to minimum value
+     */
     private String getMinKey(Map<String, Float> map) {
         String minKey = null;
         float minValue = Float.MAX_VALUE;
--- a/app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
@@ -158,9 +158,10 @@ public class PhotoTester {
 
     /**
      * Compare the list of ingredients extracted by OCR and the correct list of ingredients
-     * @param correct correct list of ingredients loaded from file
-     * @param extracted list of ingredients extracted by the OCR
-     * @return confidence percentage based on number of matched words, their similarity and order
+     *
+     * @param correct       Correct list of ingredients loaded from file
+     * @param extracted     List of ingredients extracted by the OCR
+     * @return Confidence percentage based on number of matched words, their similarity and order
      * @author Francesco Pham credit to Stefano Romanello for Levenshtein library suggestion
      */
     private float ingredientsTextComparison(String correct, String extracted){
@@ -174,11 +175,11 @@ public class PhotoTester {
         Log.i(TAG, "ingredientsTextComparison -> Start of comparing");
         Log.i(TAG, "ingredientsTextComparison -> correctWords.length == " + correctWords.length + ", extractedWords.length == " + extractedWords.length);
 
-        float points = 0;
-        int maxPoints = 0;
-        int posLastWordFound = 0;
-        int consecutiveNotFound = 0;
-
+        float points = 0; //points are added each time a word is found
+        int maxPoints = 0; //maximum points which is the number of characters of all words with more than 3 characters
+        int posLastWordFound = 0; //index of the last word found
+        int consecutiveNotFound = 0; //consecutive words not found since last word found
+        final double similarityThreshold = 0.8; //threshold above which the word we are looking at is considered found
 
         WeightedLevenshtein levenshtein = new WeightedLevenshtein(
                 new CharacterSubstitutionInterface() {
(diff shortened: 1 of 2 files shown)


app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
```diff
@@ -0,0 +1,449 @@
...
+     * @return confidence percentage based on number of matched words, their similarity and order
+     * @author Francesco Pham credit to Stefano Romanello for Levenshtein library suggestion
+     */
+    private float ingredientsTextComparison(String correct, String extracted){
+
+        extracted = extracted.toLowerCase(); //ignoring case
+
+        //split words
+        String[] extractedWords = extracted.trim().split("[ ,-:./\\n\\r]+");
+        String[] correctWords = correct.trim().split("[ ,-:./\\n\\r]+");
```
Giofasan, line 173: Thanks you made me know a function for strings that I did not know that trim() that I will insert on my function.

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
```diff
@@ -0,0 +1,449 @@
...
+                            return 0.5;
+                        }
+                        else if (c1 == 'I' && c2 == 'l') {
+                            return 0.5;
+                        }
+
+                        // For most cases, the cost of substituting 2 characters
+                        // is 1.0
+                        return 1.0;
+                    }
```
Giofasan, line 204: I advise you not to use magic numbers and then to replace them with variables defined with those numbers.

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
```diff
@@ -0,0 +1,449 @@
...
+        String[] correctWords = correct.trim().split("[ ,-:./\\n\\r]+");
+
+        Log.i(TAG, "ingredientsTextComparison -> Start of comparing");
+        Log.i(TAG, "ingredientsTextComparison -> correctWords.length == " + correctWords.length + ", extractedWords.length == " + extractedWords.length);
+
+        float points = 0;
+        int maxPoints = 0;
+        int posLastWordFound = 0;
+        int consecutiveNotFound = 0;
+
```
Giofasan: It would be useful to comment on the defined variable, writing usage

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
```diff
@@ -0,0 +1,449 @@
...
+                for (int i = 0; i < extractedWords.length && !found; i++) {
+                    //for each extracted words starting from posLastWordFound
+                    index = (posLastWordFound + i) % extractedWords.length;
+
+                    //Calculate similarity and normalize
+                    int maxLength = Math.max(word.length(),extractedWords[index].length());
+                    double similarity = 1.0 - levenshtein.distance(word,extractedWords[index])/maxLength;
+
+                    //if similarity grater than 0.8 the word is found
+                    if (similarity > 0.8) {
```
Giofasan: I do not understand why> 0.8. Is it a parameter that you gave yourself or is it default?

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
```diff
@@ -0,0 +1,449 @@
...
+                posLastWordFound = index;
+            } else {
+                consecutiveNotFound++;
+            }
+        }
+        float confidence = (points / maxPoints)*100;
+        Log.i(TAG, "ingredientsTextComparison -> confidence == " + confidence + " (%)");
+        return confidence;
+
+    }
```
Giofasan, line 267: Excellent function.
I would advise you to use less magic numbers, because you use them very often and could create problems, and replace them with local variables.

app/src/main/java/unipd/se18/ocrcamera/TestElement.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Class that contains a single test element, used in PhotoTester to build a single test and on AdapterTestElement to show data in a listview
+ * @author Luca Moroldo, Francesco Pham
+ */
+
+public class TestElement {
+
+    private Bitmap picture;
+    private JSONObject jsonObject;
```
Giofasan, line 16: I did not know the JSONObject object, I did a search on the internet and it seems very useful.
A little advice for the future: if you use new objects, add the link where the functions of the objects are explained, to simplify the review.

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
frankplus: It doesn't make sense to replace them with final variables and define them elsewhere in the code because that is where I am overriding the cost function for the Levenshtein algorithm.

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
frankplus: commented in commit 56d3cb4

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
frankplus: I made it a final variable in commit 56d3cb4

app/src/main/java/unipd/se18/ocrcamera/TestElement.java
frankplus: thank you for your advice. JSONObject is available in the standard JDK so there are loads of documentations available on the web

PR #54, opened 2018-12-01
CodeReview of 14 from 6

frankplus:
Hello Bullaku,
here is some of my code that needs to be reviewed.
could you please review:
in PhotoTester class
1) float ingredientsTextComparison(String correct, String extracted);
2) String getTagsStatsString()
3) String getMinKey(Map<String, Float> map)

and TestElement Class
Thanks

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
@@ -429,6 +429,11 @@ public class PhotoTester {
         return report;
     }
 
+    /**
+     * Find and return key corresponding to minimum value
+     * @param map
+     * @return key corresponding to minimum value
+     */
     private String getMinKey(Map<String, Float> map) {
         String minKey = null;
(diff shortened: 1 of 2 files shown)


app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+    public TestElement[] getTestElements() {
+        return testElements.toArray(new TestElement[0]);
+    }
+
+
+    /**
+     * Compare the list of ingredients extracted by OCR and the correct list of ingredients
+     * @param correct correct list of ingredients loaded from file
```
TaulantBullaku: I would have distinguished in a clearer way the "correct" parameter from the sentence that explains it otherwise it seems a repetition

frankplus: Corrected in commit 4ca6f2f

PR #55, opened 2018-12-01
Codereview of 7 from 14

GiovanniFurlan:
Hi Pham,
I'm asking for a review of the class JSONReportParser.
l added class TestEntry because it is essencial to understand the class.
No review of TestEntry class requested.

The informations about the class are the comments in the class


app/src/main/java/unipd/se18/ocrcamera/JSONReportParser.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import static android.support.constraint.Constraints.TAG;
+
+//All methods are static so no need for an dedicated class, i made it to have a clearer code while im writing it
+//The methods have to be added to the class that calls them
+//The class refers to TestEntry class
+//The class can receive only a particular type of JSONObject to have a idea see the example at this link:
+// https://ingsw18.slack.com/archives/CDFU418Q4/p1541678187026900
```
frankplus: When you say 
> All methods are static so no need for an dedicated class

do you mean a dedicated **object**?

Could you explain what do you mean by:
> The methods have to be added to the class that calls them

You can also make these multiline comments using /* */

app/src/main/java/unipd/se18/ocrcamera/JSONReportParser.java
```diff
@@ -0,0 +1,116 @@
...
+
+public class JSONReportParser {
+
+
+    /**
+     * Convert a string in JSONObject and call metod JSONReader
```
frankplus: The string is not converted into a JSONObject but into an array of TestEntry

frankplus: The method JSONReader is not called

app/src/main/java/unipd/se18/ocrcamera/JSONReportParser.java
```diff
@@ -0,0 +1,116 @@
...
+        {
+            Log.e(TAG, "Error in JSONReader");
+        }
+        return entryArray;
+    }
+
+    /**
+     * Add all the given JSONObject value in a TestEntry
+     * @param json the JSONObject where to find the value
+     * @return TestEntry with all the JSONObject value
```
frankplus, line 83: what happens if the json object doesn't have the fields "original_name" or "confidence"or  "ingredients" or "notes"?

app/src/main/java/unipd/se18/ocrcamera/JSONReportParser.java
```diff
@@ -0,0 +1,116 @@
...
+        try
+        {
+            //Iterator that slides all JSONObject inside the param
+            Iterator<String> keys = json.keys();
+            int i = 0;
+            while(keys.hasNext())
+            {
+                //Add all the information inside a TestEntry
+                JSONObject jsonDetails = json.getJSONObject(keys.next());
+                TestEntry entry = extractValues(jsonDetails);
```
frankplus, line 68: null value could be returned?

app/src/main/java/unipd/se18/ocrcamera/JSONReportParser.java
GiovanniFurlan: For the first one i mean that you can just delete this class and copy all the methods in the class where those are used and all work the same way because those are static methods.
For the comments yes i know. I used the shortcut Ctrl+/ to comment the line where u are.

app/src/main/java/unipd/se18/ocrcamera/JSONReportParser.java
GiovanniFurlan: The method JSONReader is not called because i didn't import the rest of the class

app/src/main/java/unipd/se18/ocrcamera/JSONReportParser.java
GiovanniFurlan: the method don't work, i have to add that on the specifics, because the JSONObject given have always those fields

app/src/main/java/unipd/se18/ocrcamera/JSONReportParser.java
GiovanniFurlan: The method return null only if u give this method a JSONObject with no fields "original_name" and "confidence". I had it to specifics.

GiovanniFurlan:
I made a new commit with the changes, please check if it's all ok. Thank you.

app/src/main/java/unipd/se18/ocrcamera/JSONReportParser.java
frankplus: I don't think that adding that to the specifics is a good idea because the programmer cannot always guarantee that. What is given to the method depends on the json file stored in the memory and we don't know if all json files are well formatted. I think it would be a better idea to leave the corresponding value empty (e.g. if "notes" is not present then do not set any notes in the entry).

GiovanniFurlan: The method JSONObject.getString() throws an exception if no mapping name is found, the exception is caught. I think it's good to advice the reader that the method don't work properly with different kind of JSONObject. What do you think?

frankplus: I think the method should manage those cases better because the programmer who use the method cannot guarantee that the json object have all the fields "original_name","confidence", "ingredients", "notes" and "tags".  If the json have a missing field then all the rest of TestEntry would be thrown away. Maybe you could put a try catch block each time you use json.getString and manage those cases separately. This is just my opinion.

GiovanniFurlan: Thanks for the advice but I think that this way the code would be not clear and difficult to read

frankplus: ok

PR #56, opened 2018-12-01
Codereview to 7 from 20

GiovanniFurlan:
Hi Vicentini,
I'm asking for a review of the class JSONReportParser.
l added class TestEntry because it is essencial to understand the class.
No review of TestEntry class requested.

The information about the class are in the comments of the code.


app/src/main/java/unipd/se18/ocrcamera/JSONReportParser.java
```diff
@@ -0,0 +1,116 @@
...
+
+import android.util.Log;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import static android.support.constraint.Constraints.TAG;
+
```
RobertoUnipd18, line 11: All clear and simply the only one is the indentation of the brackets but it is subjective so we wait for the approval of the style document

GiovanniFurlan: ok, thanks for the tip.

PR #57, opened 2018-12-01
Codereview to 0 from 7

Giofasan:
Hi Giovanni Furlan,
here are the parts of my code that should be revised.
the only class of OcrCamera is the ResultActivity, in which you will find two functions made with Giovanni Piva and a function for the gallery. You will find everything commented with the author.
Since there is not a lot of code to revise, I have also added three other classes of mine: OCRActivity, OcrManager, MainApplication which were of the app with tesseract.

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
@@ -11,7 +11,11 @@ import java.io.OutputStream;
 
 public class MainApplication extends Application {
 
-    public static MainApplication instance = null;
+    private static MainApplication instance = null;
+
+    private String fileName = "ita.traineddata";
+
+    private int dim = 1024;
 
     public void onCreate(){
         super.onCreate();
@@ -55,17 +59,17 @@ public class MainApplication extends Application {
                 OutputStream out = null;
                 try{
                     Log.d("MainApplication", "CopyTessDataForTextRecognizor");
-                    InputStream in = assetManager.open("ita.traineddata");
+                    InputStream in = assetManager.open(fileName);
                     String tesspath = instance.tessDataPath();
                     File tessFolder = new File(tesspath);
                     if(!tessFolder.exists()) {
                         tessFolder.mkdir();
                     }
-                    String tessData = tesspath+"/"+"ita.traineddata";
+                    String tessData = tesspath+"/"+fileName;
                     File tessFile = new File(tessData);
                     if(!tessFile.exists()){
                         out = new FileOutputStream(tessData);
-                        byte[] buffer = new byte[1024];
+                        byte[] buffer = new byte[dim];
                         int read = in.read(buffer);
                         while (read != -1){
                             out.write(buffer, 0, read);
@@ -85,7 +89,7 @@ public class MainApplication extends Application {
                             out.close();
                         }
                     }catch(Exception exx){
-
+                        Log.d("MainApplication", " Error ");
                     }
                 }
             }
--- a/app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
@@ -1,0 +1,96 @@
+package com.example.mattia.fotocamera;
+
+import android.app.Application;
+import android.content.res.AssetManager;
+import android.util.Log;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+public class MainApplication extends Application {
+
+    public static MainApplication instance = null;
+
+    public void onCreate(){
+        super.onCreate();
+        //start copy file here
+        instance=this;
+        copyTessDataForTextRecognizor();
+    }
+
+    /**
+   	* @author Fasan Giovanni
+    * @param none
+   	* @return String of tess's data path
+    */
+
+    private String tessDataPath(){
+        return MainApplication.instance.getExternalFilesDir(null)+"/tessdata/";
+    }
+
+    /**
+    * @author Fasan Giovanni
+    * @param none
+    * @return String of tess's data parent directory
+    */
+
+    public String getTessDataParentDirectory(){
+        return MainApplication.instance.getExternalFilesDir(null).getAbsolutePath();
+    }
+
+    /**
+    * @author Fasan Giovanni
+    * @param none
+    * @return void
+    * open and read tessdata file for recognizer the text
+    */
+
+    private void copyTessDataForTextRecognizor(){
+        Runnable run = new Runnable() {
+            @Override
+            public void run() {
+                AssetManager assetManager = MainApplication.instance.getAssets();
+                OutputStream out = null;
+                try{
+                    Log.d("MainApplication", "CopyTessDataForTextRecognizor");
+                    InputStream in = assetManager.open("ita.traineddata");
+                    String tesspath = instance.tessDataPath();
+                    File tessFolder = new File(tesspath);
+                    if(!tessFolder.exists()) {
+                        tessFolder.mkdir();
+                    }
+                    String tessData = tesspath+"/"+"ita.traineddata";
+                    File tessFile = new File(tessData);
+                    if(!tessFile.exists()){
+                        out = new FileOutputStream(tessData);
+                        byte[] buffer = new byte[1024];
+                        int read = in.read(buffer);
+                        while (read != -1){
+                            out.write(buffer, 0, read);
+                            read = in.read(buffer);
+                        }
+                        Log.d("MainApplication", " Did finish copy tess file ");
+                    }
+                    else{
+                        Log.d("MainApplication", " tess file exist ");
+                    }
+                }catch(Exception e){
+                    Log.d("MainApplication", "couldn't copy with the following error :"+e.toString());
+                }
+                finally{
+                    try{
+                        if(out!=null){
+                            out.close();
+                        }
+                    }catch(Exception exx){
+
+                    }
+                }
+            }
+        };
+        new Thread(run).start();
+    }
+
+}
--- a/app/src/main/java/unipd/se18/ocrcamera/OCRActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/OCRActivity.java
@@ -1,0 +1,38 @@
+package com.example.mattia.fotocamera;
+
+import android.content.Intent;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.support.v7.app.AppCompatActivity;
+import android.os.Bundle;
+import android.util.Log;
+import android.widget.ImageView;
+import android.widget.TextView;
+
+import java.io.File;
+
+/**
+* @author Fasan Giovanni
+*/
+public class OCRActivity extends AppCompatActivity {
+    //private static final String TAG = "AndroidCameraApi";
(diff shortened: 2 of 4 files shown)


app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -0,0 +1,303 @@
...
+     * @author Giovanni Fasan(g1), Giovanni Piva(g1)
+     * @param text String in which you have to find the ingredients
+     * @return String with the list of ingredients found in the text
+     * Put in database csv: https://www.youtube.com/watch?v=i-TqNzUryn8
+     */
+    public String inciDetector(String text){
+        String inci = "";
+        try {
+            InputStream is = getResources().openRawResource(R.raw.database);
+            BufferedReader reader = new BufferedReader(new InputStreamReader(is, Charset.forName("UTF-8")));
```
GiovanniFurlan: i didn't understand what is "UTF-8". 
If it's a name could be better to define it as a final variable to avoid HardCoding.

The same problem in line 273

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -0,0 +1,303 @@
...
+        canvas.drawBitmap(original, 5.0f, 0.0f, null);
+
+        return rotatedBitmap;
+    }
+
+    /**
+     * @author Giovanni Fasan(g1), Giovanni Piva(g1)
+     * @param text String in which you have to find the ingredients
+     * @return String with the list of ingredients found in the text
+     * Put in database csv: https://www.youtube.com/watch?v=i-TqNzUryn8
```
GiovanniFurlan: It's really good to link a video but it's not comfortable to watche a 20 minute video to understand a 30 lines method.
Could be better to add a comment to explain what method do.

The same problem in line 263

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -0,0 +1,303 @@
...
+
+        return inci;
+    }
+
+    /**
+     * @author Giovanni Fasan(g1), Giovanni Piva(g1)
+     * @param text String in which you have to find the ingredients
+     * @return String with ingredients list of the label
+     * Put in database csv: https://www.youtube.com/watch?v=i-TqNzUryn8
+     * problem contains
```
GiovanniFurlan: What does "problem contains" mean, i don't understand.

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -0,0 +1,303 @@
...
+
+        Bitmap rotatedBitmap = Bitmap.createBitmap(original, 0, 0, width, height, matrix, true);
+        Bitmap mutableBitmap = rotatedBitmap.copy(Bitmap.Config.ARGB_8888, true);
+        Canvas canvas = new Canvas(mutableBitmap);
+        canvas.drawBitmap(original, 5.0f, 0.0f, null);
+
+        return rotatedBitmap;
+    }
+
+    /**
```
GiovanniFurlan: There are no comments inside both methods so it's hard do understand every line of code.
Could be better to add more comments, mostly when initialize some object or use method. 
For example line 244-245, cicle in line 247, 272, 285.

GiovanniFurlan (changes requested):
sorry this was a mistake, ignore this comment

app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
```diff
@@ -0,0 +1,96 @@
...
+
+    private void copyTessDataForTextRecognizor(){
+        Runnable run = new Runnable() {
+            @Override
+            public void run() {
+                AssetManager assetManager = MainApplication.instance.getAssets();
+                OutputStream out = null;
+                try{
+                    Log.d("MainApplication", "CopyTessDataForTextRecognizor");
+                    InputStream in = assetManager.open("ita.traineddata");
```
GiovanniFurlan: Is "ita.traineddata" a String you chose? To avoid hardcoding could be better to define a final variable at the beginning of the method.

app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
```diff
@@ -0,0 +1,96 @@
...
+                    String tesspath = instance.tessDataPath();
+                    File tessFolder = new File(tesspath);
+                    if(!tessFolder.exists()) {
+                        tessFolder.mkdir();
+                    }
+                    String tessData = tesspath+"/"+"ita.traineddata";
+                    File tessFile = new File(tessData);
+                    if(!tessFile.exists()){
+                        out = new FileOutputStream(tessData);
+                        byte[] buffer = new byte[1024];
```
GiovanniFurlan: Could be better use a final variable with value 1024 to avoid hardcoding.

app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
```diff
@@ -0,0 +1,96 @@
...
+                        while (read != -1){
+                            out.write(buffer, 0, read);
+                            read = in.read(buffer);
+                        }
+                        Log.d("MainApplication", " Did finish copy tess file ");
+                    }
+                    else{
+                        Log.d("MainApplication", " tess file exist ");
+                    }
+                }catch(Exception e){
```
GiovanniFurlan, line 83: Could be better to not include general Exception but only the one refers to this try and catch

app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
```diff
@@ -0,0 +1,96 @@
...
+    * @author Fasan Giovanni
+    * @param none
+    * @return String of tess's data parent directory
+    */
+
+    public String getTessDataParentDirectory(){
+        return MainApplication.instance.getExternalFilesDir(null).getAbsolutePath();
+    }
+
+    /**
```
GiovanniFurlan, line 47: For me that i never used Tesseract was really difficult to understand every method of this library.
Could be better to use comments inside the code when you use method of this library so would be easier to me to understabnd

app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
```diff
@@ -0,0 +1,96 @@
...
+                    }
+                }catch(Exception e){
+                    Log.d("MainApplication", "couldn't copy with the following error :"+e.toString());
+                }
+                finally{
+                    try{
+                        if(out!=null){
+                            out.close();
+                        }
+                    }catch(Exception exx){
```
GiovanniFurlan, line 91: Could be better to not call a void catch, difficult to find a bug this way, or to add a comment that explain why u do this.
Also not to catch general Exception

app/src/main/java/unipd/se18/ocrcamera/OCRActivity.java
```diff
@@ -0,0 +1,38 @@
...
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_ocr);
+        OcrManager manager = new OcrManager();
+        //initialize OCR library
+        manager.initAPI();
+        extras = getIntent().getExtras();
+        //get a photo's path from MainActivity
+        String path=extras.getString("PATH_I_NEED");
```
GiovanniFurlan, line 29: Is "PATH_I_NEED" a String you chose? To avoid hardcoding could be better to define a final variable at the beginning of the method.

app/src/main/java/unipd/se18/ocrcamera/OcrManager.java
```diff
@@ -0,0 +1,36 @@
...
+    public void  initAPI(){
+        baseAPI = new TessBaseAPI();
+        String dataPath = MainApplication.instance.getTessDataParentDirectory();
+        baseAPI.init(dataPath,"ita");
+        //first param is datapath which is part to the your trainned data, second is language code
+        //now, your trained data stored in asset folder, we need to copy it to another exernal storage folder
+        //It is better do this work when application start firt time
+    }
+
+    /**
```
GiovanniFurlan, line 23: Could be better to add @modify because the method call setImage.

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -299,5 +320,42 @@ public String inciDetectorEtichetta(String text){
...
+                mImageView.setImageBitmap(Bitmap.createScaledBitmap(lastPhoto, lastPhoto.getWidth(), lastPhoto.getHeight(), false));
+                AsyncLoad ocrTask = new AsyncLoad(mOCRTextView,getString(R.string.processing));
+                ocrTask.execute(lastPhoto);
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+        }
+    }
+
 }
```
GiovanniFurlan, line 361: I cannot comment all the line so i made all the other comments here:
-line 254 and 278  Could be better to add a comment to the method that explains what it does.

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -299,5 +320,42 @@ public String inciDetectorEtichetta(String text){
...
+    protected void onActivityResult(int requestCode, int resultCode, Intent data){
+        super.onActivityResult(requestCode,resultCode,data);
+        if(resultCode==RESULT_OK && requestCode==PICK_IMG){
+            Uri imageUri;
+            imageUri = data.getData();
+            try {
+                lastPhoto = MediaStore.Images.Media.getBitmap(this.getContentResolver(), imageUri);
+                mImageView = findViewById(R.id.img_captured_view);
+                mImageView.setImageBitmap(Bitmap.createScaledBitmap(lastPhoto, lastPhoto.getWidth(), lastPhoto.getHeight(), false));
+                AsyncLoad ocrTask = new AsyncLoad(mOCRTextView,getString(R.string.processing));
```
GiovanniFurlan: I didn't understand the line 352, no information on internet. Could be better to add a comment to explain that.

GiovanniFurlan (changes requested):
Hi Fasan,
I reviewed your code. 
I had some problem on the code with few comments, instead the code with a lot of comments was clear.

app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
Giofasan: Yes you're right it's actually hardcoding.
ita.traineddata is the name of the library file.

app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
Giofasan: As we have discarded Tesseract, I preferred to focus on something else and I have not properly commented on the code.
For this I apologize.

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
Giofasan: I used the same method previously written by someone else in the class. That's why I did not add any comments. I will try to add the comment.

Giofasan:
I have corrected some suggested things.
Now do you think it's better?

GiovanniFurlan (approved):

PR #58, opened 2018-12-01
Codereview of 3 From 10

gioppiva:
Hi, you can find my code in  "CodeReviewGioPiva".

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
@@ -1,0 +1,300 @@
+package unipd.se18.ocrcamera;
+
+import android.annotation.SuppressLint;
+import android.app.ProgressDialog;
+import android.content.Context;
+import android.content.Intent;
+import android.content.SharedPreferences;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.graphics.Canvas;
+import android.graphics.Matrix;
+import android.media.Image;
+import android.os.AsyncTask;
+import android.os.Bundle;
+import android.os.Environment;
+import android.support.design.widget.FloatingActionButton;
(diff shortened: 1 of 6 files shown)


app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -0,0 +1,301 @@
...
+        try {
+            InputStream is = getResources().openRawResource(R.raw.database);
+            BufferedReader reader = new BufferedReader(new InputStreamReader(is, Charset.forName("UTF-8")));
+            String ingredient;
+            while ((ingredient=reader.readLine())!=null){
+                if (text.toUpperCase().contains(ingredient)) {
+                    inci = inci + "<b>" + ingredient + "</b>; ";
+                }
+            }
+        }catch (IOException e){
```
AlbVal97, line 252: I don't understand why our Style document about braces are always followed with the exception of this line: I suggest to leave a space between "{" and "catch".

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -0,0 +1,301 @@
...
+                    }
+                }
+                is.reset();
+                if(!listInci.contains(box)) {
+                    listInci.add(box);
+                }
+            }
+
+            for(int i=0; i<listInci.size(); i++){
+                inci = inci + "<b>" + listInci.get(i) + "</b>; "; /**bolding ingredients founded*/
```
AlbVal97: It's likely that you intended "found" instead of "founded" in the comment above

AlbVal97 (commented):
Good work overall, but I have to point out the fact that our Style document has been not so followed in this fragment. Specifically spacing between binary and assignment operators has been widely misused, in addition to other minor imperfections.

PR #59, opened 2018-12-01
CodeReview of 0 from 13

Giofasan:
Hi Luca Moroldo,
here are the parts of my code that should be reviewed.
the only class of OcrCamera is the ResultActivity, in which you will find two functions made with Giovanni Piva and a function for the gallery. You will find everything commented with the author.
Since there is not much code to be reviewed I added three more my classes: OCRActivity, OcrManager, MainApplication who were of the app with tesseract.

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
@@ -11,7 +11,11 @@ import java.io.OutputStream;
 
 public class MainApplication extends Application {
 
-    public static MainApplication instance = null;
+    private static MainApplication instance = null;
+
+    private String fileName = "ita.traineddata";
+
+    private int dim = 1024;
 
     public void onCreate(){
         super.onCreate();
@@ -55,17 +59,17 @@ public class MainApplication extends Application {
                 OutputStream out = null;
                 try{
                     Log.d("MainApplication", "CopyTessDataForTextRecognizor");
-                    InputStream in = assetManager.open("ita.traineddata");
+                    InputStream in = assetManager.open(fileName);
                     String tesspath = instance.tessDataPath();
                     File tessFolder = new File(tesspath);
                     if(!tessFolder.exists()) {
                         tessFolder.mkdir();
                     }
-                    String tessData = tesspath+"/"+"ita.traineddata";
+                    String tessData = tesspath+"/"+fileName;
                     File tessFile = new File(tessData);
                     if(!tessFile.exists()){
                         out = new FileOutputStream(tessData);
-                        byte[] buffer = new byte[1024];
+                        byte[] buffer = new byte[dim];
                         int read = in.read(buffer);
                         while (read != -1){
                             out.write(buffer, 0, read);
@@ -85,7 +89,7 @@ public class MainApplication extends Application {
                             out.close();
                         }
                     }catch(Exception exx){
-
+                        Log.d("MainApplication", " Error ");
                     }
                 }
             }
--- a/app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
@@ -1,0 +1,96 @@
+package com.example.mattia.fotocamera;
+
+import android.app.Application;
+import android.content.res.AssetManager;
+import android.util.Log;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+public class MainApplication extends Application {
+
+    public static MainApplication instance = null;
+
+    public void onCreate(){
+        super.onCreate();
+        //start copy file here
+        instance=this;
+        copyTessDataForTextRecognizor();
+    }
+
+    /**
+   	* @author Fasan Giovanni
+    * @param none
+   	* @return String of tess's data path
+    */
+
+    private String tessDataPath(){
+        return MainApplication.instance.getExternalFilesDir(null)+"/tessdata/";
+    }
+
+    /**
+    * @author Fasan Giovanni
+    * @param none
+    * @return String of tess's data parent directory
+    */
+
+    public String getTessDataParentDirectory(){
+        return MainApplication.instance.getExternalFilesDir(null).getAbsolutePath();
+    }
+
+    /**
+    * @author Fasan Giovanni
+    * @param none
+    * @return void
+    * open and read tessdata file for recognizer the text
+    */
+
+    private void copyTessDataForTextRecognizor(){
+        Runnable run = new Runnable() {
+            @Override
+            public void run() {
+                AssetManager assetManager = MainApplication.instance.getAssets();
+                OutputStream out = null;
+                try{
+                    Log.d("MainApplication", "CopyTessDataForTextRecognizor");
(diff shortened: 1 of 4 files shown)


app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
```diff
@@ -0,0 +1,96 @@
...
+import android.util.Log;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+public class MainApplication extends Application {
+
+    public static MainApplication instance = null;
```
lucamoroz: The access to this instance can be private, anyway you can just remove it and call methods like getExternalFilesDir() and getAssets() directly (line 30, 40 and 54) as they belong to Context class. 
If you need to get the external files dir from outside the application (line 16 - OcrManager) you may consider passing as argument the context and use it to call getExternalFilesDir() or creating an appropriate static string variable to embrace OOP encapsulation.

app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+    * @author Fasan Giovanni
+    * @param none
+    * @return void
+    * open and read tessdata file for recognizer the text
+    */
+
+    private void copyTessDataForTextRecognizor(){
+        Runnable run = new Runnable() {
```
lucamoroz, line 55: Doing this kind of tasks in background is a good style! But be aware that if the files you are reading/writing are too big, you may have troubles when you init the OCR (line 14 - OcrManager) and this task ain't finished

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -0,0 +1,361 @@
...
+     * @return String with the list of ingredients found in the text
+     * Put in database csv: https://www.youtube.com/watch?v=i-TqNzUryn8
+     */
+    public String inciDetector(String text){
+        String inci = "";
+        try {
+            InputStream is = getResources().openRawResource(R.raw.database);    //InputStream from database.csv
+            BufferedReader reader = new BufferedReader(new InputStreamReader(is, Charset.forName("UTF-8")));    //database reader
+            String ingredient;  //String of inci ingredient
+            while ((ingredient=reader.readLine())!= null){      //defines ingredient as each row of the database
```
lucamoroz, line 265: In this way you scan all the database each time, this operation takes O(n) where n is the size of the database (plus the contains() cost). You can reduce the number of operations by splitting the text given in input into words and check for each word if there is an ingredient with this word in the database; if the ingredients are indexed this would take less as the search wouldn't be sequential.

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -0,0 +1,361 @@
...
+            String box="";
+
+            for (int i=0; i<tokens.length; i++) {
+                while ((ingredient=reader.readLine())!=null){       //defines ingredient as each row of the database
+                    if (tokens[i].toUpperCase().contains(ingredient)) {         //Use .toUpperCase because the database's words are uppercase. .contains find if tokens[i] is an ingredient
+                        box=ingredient;     //save the ingredient in a box. after while i use the last ingredient that I found
+                    }
+                }
+                is.reset();     //reset InputStream to the mark 0
+                if (!listInci.contains(box)) {
```
lucamoroz, line 305: If you don't want duplicates, you may consider using a set data structure that by default checks for duplicates when you try to add an item to the set.

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
```diff
@@ -0,0 +1,361 @@
...
+                    listInci.add(box);          //check if there are duplicates
+                }
+            }
+
+            if (listInci.size()>1) {
```
lucamoroz, line 310: I didn't understand: why are you checking if the size is >1 and not >0?

lucamoroz (commented):
I've added some inline comments.
May I kindly ask you to add some more explanation comments in future, so I'll be able to easily comprehend  the context next time. Thank you very much.

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
Giofasan: I know very well that this function has the very bad performance, I used it mainly as a base to write inciDetectorEtichetta

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
Giofasan: Excellent thanks a lot for the advice, we worked on it

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
Giofasan: Even I do not understand why I put 1 (lol), probably without thinking I put 1 and then I have not corrected. I correct immediately.

app/src/main/java/unipd/se18/ocrcamera/ResultActivity.java
lucamoroz: Ok, I think that these kind of optimizations can also be implemented at the "final phase", i.e. after being sure of the correctness of the method

app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
```diff
@@ -0,0 +1,96 @@
...
+                }catch(Exception e){
+                    Log.d("MainApplication", "couldn't copy with the following error :"+e.toString());
+                }
+                finally{
+                    try{
+                        if(out!=null){
+                            out.close();
+                        }
+                    }catch(Exception exx){
+
```
lucamoroz: You shouldn't leave an empty catch here, and also at line 113.

lucamoroz (changes requested):
I noticed you left some empty catch in the code, please fill them with an error handling or a log (line 87 and 113).

app/src/main/java/unipd/se18/ocrcamera/MainApplication.java
```diff
@@ -0,0 +1,96 @@
...
+    private void copyTessDataForTextRecognizor(){
+        Runnable run = new Runnable() {
+            @Override
+            public void run() {
+                AssetManager assetManager = MainApplication.instance.getAssets();
+                OutputStream out = null;
+                try{
+                    Log.d("MainApplication", "CopyTessDataForTextRecognizor");
+                    InputStream in = assetManager.open("ita.traineddata");
```
lucamoroz: Please don't hard-code the strings, can you add a variable for that? (You have done the same in many lines)

lucamoroz (changes requested):
I noticed some hard-coded strings, can you please replace them with a variable or a string resource?

Giofasan:
I have corrected some suggested things.
Now do you think it's better?

lucamoroz:
> I have corrected some suggested things.
> Now do you think it's better?

Typically when you define a constant string you use the keyword 'final' and name the variable with a capitalized word, for example: private static final String TRAINED_DATA_FILENAME = "ita.traineddata"

PR #62, opened 2018-12-01
CodeReview of 13 from 20

lucamoroz:
Dear Mr. Vicentini, 
I hope this message finds you well. 
Please find below the files containing the code that I would kindly ask you to review. 
The methods involved in this review are inside PhotoTester class:
   - PhotoTester
   - testAndReport
   - writeReportToExternalStorage
   - class RunnableTest
   - addTestElement

Please be aware that I've also added two files that don't need any kind of revision, as for now class TestElement is tested by class TestElementTest (using jUnit), but you should be aware of the available methods.

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
@@ -64,10 +64,14 @@ public class PhotoTester {
             alteration16.jpg
             alteration22.jpg
 
-         To find out which 'foto' has an alteration (e.g. modified photo) we need to read each .txt file, data is stored in JSON format.
+         A single test has an image and a description written in JSON format that contains tags, ingredients in the image, and notes.
+         A single test can also have some edited images associated to it, in this case the description also contains each edited image
+         data (image file name, notes, tags).
+         To find out which 'foto' has an alteration (e.g. modified photo) we need to read each .txt file.
 
          To automate the elaboration of each test i've modified the class TestElement, that allows to link each photo to its description and alterations (if any),
          and then, at the evaluation phase, set the results like recognition confidence and recognized text.
+         Confidence is a float that describes the correspondence between the recognized text from the OCR and the real one.
          */
 
         //Credits for getStorageDir function: Pietro Prandini (group 2)
--- a/app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
@@ -1,0 +1,306 @@
+package unipd.se18.ocrcamera;
+
+import android.graphics.Bitmap;
+import android.util.Log;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.io.File;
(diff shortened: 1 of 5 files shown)


app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
```diff
@@ -0,0 +1,306 @@
...
+            - FixedThreadPool: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Executors.html#newFixedThreadPool-int-
+            - CounDownLatch: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CountDownLatch.html
+
+        When all the tasks are completed, it saves a report in JSON format and return it as String.
+         */
+        final JSONObject fullJsonReport = new JSONObject();
+
+        int totalTestElements = testElements.size();
+        //countDownLatch allows to sync this thread with the end of all the single tests
+        CountDownLatch countDownLatch = new CountDownLatch(totalTestElements);
```
RobertoUnipd18, line 153: why you  use a CountDownLatch and don't a semaphore? What are the advantages of using latch?

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
```diff
@@ -0,0 +1,306 @@
...
+                String correctIngredients = test.getIngredients();
+                //compare the text recognized by the OCR with the real one
+                float confidence = ingredientsTextComparison(correctIngredients, extractedIngredients);
+
+                //insert results in test
+                test.setConfidence(confidence);
+                test.setRecognizedText(extractedIngredients);
+
+                //evaluate each alterations if present
+                String[] alterationsFileNames = test.getAlterationsNames();
```
RobertoUnipd18, line 253: you want to test the difference between the words of ocr and the totals one??

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
lucamoroz: I could use a semaphore, but it would be more tricky to implement even if it would be a nice exercise. As there is already a component (CountDownLatch) that do what I need, I've opted for it (in this case there are no reasons to reinvent the wheel).

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
lucamoroz: I want to evaluate the performance of the OCR comparing the recognized text with the real text, the evaluation is done also for any alteration of a photo. Do you think this wasn't clear?

I added some comments at the beginning of PhotoTester constructor, I hope this clarifies the context.

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
RobertoUnipd18: ok it was just a question

RobertoUnipd18:
I have revised the code following this checklist:
Adequate comments, indentation, style document, variable declarations, code repetitions

PR #64, opened 2018-12-01
CodeReview of 16 from 8

ghost:
Here's the file for the review.
The introduction header should explain everything, but in case, fell free to PM me.
I wish you good work.

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
@@ -32,9 +32,8 @@ public class CameraActivity extends AppCompatActivity {
     //variable used to interface with the CameraView library
     private CameraView camera;
 
-    //variable (with default value false) set to true if a picture is captured
-    //only when the picture is sent to the PicturePreviewActivity the variable is set to false again
-    private boolean isPictureCaptured = false;
+    //variable used to connect to the capturePhoto button
+    private ImageButton buttonPhoto;
 
     /**
      * onCreate method of the Android Activity Lifecycle
@@ -47,7 +46,7 @@ public class CameraActivity extends AppCompatActivity {
         setContentView(R.layout.activity_camera);
         camera = findViewById(R.id.viewCamera);
 
-        //interface method for automatic onResume, onPause, onDestroy
+        //library method for automatic onResume, onPause, onDestroy
         camera.setLifecycleOwner(this);
 
         //Pinch to zoom support
@@ -60,13 +59,17 @@ public class CameraActivity extends AppCompatActivity {
         camera.addCameraListener(new CameraListener() {
             /**
              * Listening for photo taken by the camera
-             * @param picture the taken photo in array form (stream of byte)
+             * @param picture the taken photo as an array of byte (a stream of data)
              * @author Bedin Elia
              */
             public void onPictureTaken(byte[] picture) { onCapture(picture); }
         });
 
-        ImageButton buttonPhoto = findViewById(R.id.capturePhoto);
+        //connecting the variable to the button capturePhoto
+        buttonPhoto = findViewById(R.id.capturePhoto);
+        //enabling the button
+        buttonPhoto.setEnabled(true);
+
         buttonPhoto.setOnClickListener(new View.OnClickListener() {
             /**
              * Listening for button click, then call capturePhoto() to take a photo
@@ -85,18 +88,21 @@ public class CameraActivity extends AppCompatActivity {
      * @author Bedin Elia
      */
     private void capturePhoto() {
-        if (isPictureCaptured) return;
-        isPictureCaptured = true;
+        //checking if the app is ready to capture a photo (if the button is enabled)
+        if (!buttonPhoto.isEnabled()) return;
+        //if the button was enabled, set it to disabled and capture a photo
+        buttonPhoto.setEnabled(false);
         camera.capturePicture();
     }
 
     /**
-     * Method for sending the taken photo in array form to the PicturePreviewActivity
-     * @param picture the taken photo in array form (stream of byte)
+     * Method for sending the taken photo as an array of byte to the PicturePreviewActivity
+     * @param picture the taken photo as an array of byte (a stream of data)
      * @author Bedin Elia
      */
     private void onCapture(byte[] picture) {
-        isPictureCaptured = false;
+        //releasing the button, the app is ready to capture a photo again
+        buttonPhoto.setEnabled(true);
         //sending the taken photo to PicturePreviewActivity via setImage
         PicturePreviewActivity.setImage(picture);
         //sending information to PicturePreviewActivity through Intent
--- a/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
@@ -2,9 +2,9 @@
  *  The main goal for this project is to test older Camera API (15 to 19) to see if it's
  *  plausible to support these devices and lower the minimum API requirement on CommonDemo
  *  so we can support full range of Android devices (API 15 and higher is 100% demographic)
- *  I used this interface for Camera (1) API to make the program
+ *  I used this library that interfaces with the Camera (1) API to make the program
  *      https://github.com/natario1/CameraView
- *  working with master build of the interface
+ *  working with master build of the library
  *  The documentation starts here
  *      https://github.com/natario1/CameraView#usage
  */
@@ -29,8 +29,12 @@ import com.otaliastudios.cameraview.GestureAction;
  */
 public class CameraActivity extends AppCompatActivity {
 
+    //variable used to interface with the CameraView library
     private CameraView camera;
-    private boolean mCapturingPicture;
+
+    //variable (with default value false) set to true if a picture is captured
+    //only when the picture is sent to the PicturePreviewActivity the variable is set to false again
+    private boolean isPictureCaptured = false;
 
     /**
      * onCreate method of the Android Activity Lifecycle
@@ -55,11 +59,11 @@ public class CameraActivity extends AppCompatActivity {
 
         camera.addCameraListener(new CameraListener() {
             /**
-             * Listining for photo taken by the camera
-             * @param jpeg the taken photo in array form
+             * Listening for photo taken by the camera
+             * @param picture the taken photo in array form (stream of byte)
              * @author Bedin Elia
              */
-            public void onPictureTaken(byte[] jpeg) { onPicture(jpeg); }
+            public void onPictureTaken(byte[] picture) { onCapture(picture); }
         });
 
         ImageButton buttonPhoto = findViewById(R.id.capturePhoto);
@@ -74,7 +78,6 @@ public class CameraActivity extends AppCompatActivity {
                 capturePhoto();
             }
         });
-
     }
 
     /**
@@ -82,27 +85,27 @@ public class CameraActivity extends AppCompatActivity {
      * @author Bedin Elia
      */
     private void capturePhoto() {
-        if (mCapturingPicture) return;
-        mCapturingPicture = true;
+        if (isPictureCaptured) return;
+        isPictureCaptured = true;
         camera.capturePicture();
     }
 
     /**
      * Method for sending the taken photo in array form to the PicturePreviewActivity
-     * @param jpeg the taken photo
+     * @param picture the taken photo in array form (stream of byte)
      * @author Bedin Elia
      */
-    private void onPicture(byte[] jpeg) {
-        mCapturingPicture = false;
+    private void onCapture(byte[] picture) {
+        isPictureCaptured = false;
         //sending the taken photo to PicturePreviewActivity via setImage
-        PicturePreviewActivity.setImage(jpeg);
+        PicturePreviewActivity.setImage(picture);
         //sending information to PicturePreviewActivity through Intent
         Intent intent = new Intent(CameraActivity.this, PicturePreviewActivity.class);
         startActivity(intent);
     }
 
     /**
-     * Permission checker for Android M or higher
+     * Explicit permission checker for Android M or higher
      * @param requestCode permission code
      * @param permissions the requested permissions
      * @param grantResults the grant results for the corresponding permissions
--- a/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
@@ -1,0 +1,123 @@
+/*
+ *  The main goal for this project is to test older Camera API (15 to 19) to see if it's
+ *  plausible to support these devices and lower the minimum API requirement on CommonDemo
+ *  so we can support full range of Android devices (API 15 and higher is 100% demographic)
+ *  I used this interface for Camera (1) API to make the program
+ *      https://github.com/natario1/CameraView
+ *  working with master build of the interface
+ *  The documentation starts here
+ *      https://github.com/natario1/CameraView#usage
+ */
+
+package com.example.elia.compatibilitycamera;
+
+import android.content.Intent;
+import android.os.Bundle;
+import android.support.v7.app.AppCompatActivity;
+import android.view.View;
+import android.content.pm.PackageManager;
+import android.support.annotation.NonNull;
+import android.widget.ImageButton;
+
+import com.otaliastudios.cameraview.CameraListener;
+import com.otaliastudios.cameraview.CameraView;
+import com.otaliastudios.cameraview.Gesture;
+import com.otaliastudios.cameraview.GestureAction;
+
+/**
+ * Activity for camera preview and taking photos
+ */
+public class CameraActivity extends AppCompatActivity {
+
+    private CameraView camera;
+    private boolean mCapturingPicture;
+
+    /**
+     * onCreate method of the Android Activity Lifecycle
+     * @param savedInstanceState the bundle of the last instance state saved
+     * @author Bedin Elia
+     */
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_camera);
+        camera = findViewById(R.id.viewCamera);
+
+        //interface method for automatic onResume, onPause, onDestroy
+        camera.setLifecycleOwner(this);
+
+        //Pinch to zoom support
+        camera.mapGesture(Gesture.PINCH, GestureAction.ZOOM);
+        //Tap to focus support
+        camera.mapGesture(Gesture.TAP, GestureAction.FOCUS_WITH_MARKER);
+        //Long tap to capture support (currently disabled, should be .CAPTURE)
+        camera.mapGesture(Gesture.LONG_TAP, GestureAction.NONE);
+
+        camera.addCameraListener(new CameraListener() {
+            /**
+             * Listining for photo taken by the camera
+             * @param jpeg the taken photo in array form
+             * @author Bedin Elia
+             */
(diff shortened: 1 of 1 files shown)


app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,123 @@
+/*
+ *  The main goal for this project is to test older Camera API (15 to 19) to see if it's
+ *  plausible to support these devices and lower the minimum API requirement on CommonDemo
+ *  so we can support full range of Android devices (API 15 and higher is 100% demographic)
```
PrandiniUniPD, line 4: Do you mean that all the Android devices currently used have an API level equal or higher to 15? Have I understood?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,123 @@
...
+ *  I used this interface for Camera (1) API to make the program
+ *      https://github.com/natario1/CameraView
```
PrandiniUniPD, line 6: I'm sorry but maybe I've not understood, is this an interface or a library?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,123 @@
...
+ *  working with master build of the interface
+ *  The documentation starts here
+ *      https://github.com/natario1/CameraView#usage
+ */
```
PrandiniUniPD, line 10: I like this introduction because there are the aim of the work and some links about the major tools used.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,123 @@
...
+        //Pinch to zoom support
+        camera.mapGesture(Gesture.PINCH, GestureAction.ZOOM);
+        //Tap to focus support
+        camera.mapGesture(Gesture.TAP, GestureAction.FOCUS_WITH_MARKER);
+        //Long tap to capture support (currently disabled, should be .CAPTURE)
+        camera.mapGesture(Gesture.LONG_TAP, GestureAction.NONE);
+
+        camera.addCameraListener(new CameraListener() {
+            /**
+             * Listining for photo taken by the camera
```
PrandiniUniPD: Do you mean "Listening"? Have I understood?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,123 @@
...
+        });
+
+    }
+
+    /**
+     * Method for taking a photo
+     * @author Bedin Elia
+     */
+    private void capturePhoto() {
+        if (mCapturingPicture) return;
```
PrandiniUniPD: Should be a good idea to do this disabling the button? Similarly to `buttonPhoto.setEnabled(false);`

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Permission checker for Android M or higher
+     * @param requestCode permission code
+     * @param permissions the requested permissions
+     * @param grantResults the grant results for the corresponding permissions
+     * @author Bedin Elia
+     */
+    @Override
+    public void onRequestPermissionsResult
```
PrandiniUniPD, line 119: I have not clear where in the code there is the checking of the permissions that eventually launch this method.
Accordingly to this https://developer.android.com/training/permissions/requesting#perm-check it should be similar to this:
`if (ContextCompat.checkSelfPermission(CameraActivity.this, Manifest.permission.CAMERA)`
`        != PackageManager.PERMISSION_GRANTED) {`
`    ActivityCompat.requestPermissions(CameraActivity.this,` 
`                new String[]{Manifest.permission.CAMERA},`
`                MY_PERMISSIONS_REQUEST_CAMERA);`
`}`
Have I understood your intentions? Is it included in the library so it's not necessary?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,123 @@
...
+import com.otaliastudios.cameraview.Gesture;
+import com.otaliastudios.cameraview.GestureAction;
+
+/**
+ * Activity for camera preview and taking photos
+ */
+public class CameraActivity extends AppCompatActivity {
+
+    private CameraView camera;
+    private boolean mCapturingPicture;
```
PrandiniUniPD: Is it a good idea to explicitly set this variable to false and put a short description about the future use of this?
But firstly is it a good idea to disable and enable the functionality of the button instead a boolean system?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /**
+     * Method for taking a photo
+     * @author Bedin Elia
+     */
+    private void capturePhoto() {
+        if (mCapturingPicture) return;
+        mCapturingPicture = true;
+        camera.capturePicture();
```
PrandiniUniPD, line 93: I'll try to explain you what I've understood (thanks in advance for the support):
This statement launches the callback for taking a picture and so the method onPictureTaken (line 62) is launched. Finally the onPictureTaken launches the onPicture method (line 95) that firstly sets the boolean variable mCapturingPicture to false. Is it right?
If mCapturingPicture is not already setted to false the capturePhoto method doesn't launch the camera.capturePicture() sentence but returns at the line 85. Is it right?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,123 @@
...
+             * Listening for button click, then call capturePhoto() to take a photo
+             * @param view the layout components
+             * @author Bedin Elia
+             */
+            @Override
+            public void onClick(View view) {
+                capturePhoto();
+            }
+        });
+
```
PrandiniUniPD: Is this empty line permitted here for our Document Style?

PrandiniUniPD (commented):
Hi Elia, I've reviewed your code, thanks in advance for your patience, maybe I've not understood everything. So I've commented which are my observations, we could speak about these and finally have a common approved version of your code.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: Yes. When creating a new project in Android Studio you have to set the minimum API requirement for your app; with 15 and higher you are targeting all (it's actually 99.8%, but Android Studio approximates to 100%  and I don't really get why since out of more or less 2 billion Android devices, **2% is still 40000000 devices with API lower than 15**) Android devices.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: Sorry, it is a bit misleading. It's an Android library that interfaces with Camera API.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: Yes, it's a typo, thank you for pointing out!

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: ImageButton does not have a setEnabled, but a android:clickable="false" XML tag as far as I know.
I'll give it a shot soon, thanks you!

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: The library already set the permission for the camera, but it's implicit like in Android K (and lower), when there was no prompt to ask you.
This code it's for having a visual prompt, and it's an override for the current library method.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: You are absolutely right, thank you.
I'll add descriptions and set the variable mCapturingPicture to default 'false'.

> But firstly is it a good idea to disable and enable the functionality of the button instead a boolean system?

See https://github.com/PrandiniUniPD/OCRCamera/pull/64#discussion_r238099864

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: You are correct, sorry if it was hard to understand.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: Another typo, thank you for pointing out.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
PrandiniUniPD: Right, for do this by a xml file but you use an ImageButton object programmatically.
An ImageButton object extends ImageView and ImageView extends View.
In the View object you can enable or disable the state of a View (or a subclassed object, same as ImageButton) programmatically
https://developer.android.com/reference/android/view/View#setEnabled(boolean) 
Could be useful?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
PrandiniUniPD: mmm, are you sure? Maybe I'm not understand. Is onRequestPermissionResult for showing a visual prompt for the permissions? https://developer.android.com/reference/android/support/v4/app/ActivityCompat#requestpermissions here it's not seem to be so. In that link I've read that the ActivityCompat.requestPermission launch the visual prompt to the user and the onRequestPermissionsResult catches the results of the request.
But if it is already implemented in the library it's ok.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: Please forgive my lack of experience with Android.
Your help is really appreciated, I'll read the documentation you suggested me.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: Yeah, the library does, see https://github.com/natario1/CameraView#declaration.
Thanks again.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
PrandiniUniPD: Don't worry I've proposed you only some questions and we've spoken about things, I've to learn with you. So thanks you.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
PrandiniUniPD: Ok, thanks to you for this specifications. I've left that point of the documentation! Thanks to you.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
PrandiniUniPD: Don't worry, I've asked you for knowing if I've understood the functionality. So thanks for the feedback.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
PrandiniUniPD: Ok, so if you think that it's a good idea enabling and disabling the button and you need some support ask me.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: No problem!

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: Thank you very much, I'll keep that in mind!

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,126 @@
...
+        camera.capturePicture();
+    }
+
+    /**
+     * Method for sending the taken photo in array form to the PicturePreviewActivity
+     * @param picture the taken photo in array form (stream of byte)
+     * @author Bedin Elia
+     */
+    private void onCapture(byte[] picture) {
+        isPictureCaptured = false;
```
PrandiniUniPD: Could be a good idea making the buttonPhoto a private shared variable for this class, removing isPictureCaptured and the relative statements, putting on this line (99) `buttonPhoto.setEnabled(true);` and putting in the onClick method of the line 77 `buttonPhoto.setEnabled(false)`?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,126 @@
...
+     * @author Bedin Elia
+     */
+    private void capturePhoto() {
+        if (isPictureCaptured) return;
+        isPictureCaptured = true;
+        camera.capturePicture();
+    }
+
+    /**
+     * Method for sending the taken photo in array form to the PicturePreviewActivity
```
PrandiniUniPD: In this and in the next line do you mean form or format?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,126 @@
...
+        camera.mapGesture(Gesture.PINCH, GestureAction.ZOOM);
+        //Tap to focus support
+        camera.mapGesture(Gesture.TAP, GestureAction.FOCUS_WITH_MARKER);
+        //Long tap to capture support (currently disabled, should be .CAPTURE)
+        camera.mapGesture(Gesture.LONG_TAP, GestureAction.NONE);
+
+        camera.addCameraListener(new CameraListener() {
+            /**
+             * Listening for photo taken by the camera
+             * @param picture the taken photo in array form (stream of byte)
```
PrandiniUniPD: Could you mean format instead of form?

PrandiniUniPD (commented):
Hi Elia! I've some questions for you, I hope they are useful.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: I meant "the taken photo in the form of an array (of byte)"

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: See https://github.com/PrandiniUniPD/OCRCamera/pull/64#discussion_r239197936

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
PrandiniUniPD: I've asked you because it seems that the word "form" is mostly used to indicate a shape, not a structure ( https://www.wordreference.com/enit/form ), and I thought you would like to indicate a format instead.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
PrandiniUniPD: Other explanations about my question here: https://github.com/PrandiniUniPD/OCRCamera/pull/64#discussion_r239379358

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
PrandiniUniPD: This is a proposal to avoid a boolean synchronization if it could be useful.
When the button is pressed the button is disabled and it is enabled when the processing is ended.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: Fixed in the new commit (https://github.com/PrandiniUniPD/OCRCamera/pull/64/commits/753e923d2dc820a0883776c8fc36e5c847c8a4dc).
Thanks for your help!

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: See https://github.com/PrandiniUniPD/OCRCamera/pull/64#discussion_r239414418

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: Sorry if it took me a long time to apply and test the proposed changes.
They are now applied in the new commit (https://github.com/PrandiniUniPD/OCRCamera/commit/753e923d2dc820a0883776c8fc36e5c847c8a4dc)
Thank you very much for you help on making the code better!

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,132 @@
...
+
+        buttonPhoto.setOnClickListener(new View.OnClickListener() {
+            /**
+             * Listening for button click, then call capturePhoto() to take a photo
+             * @param view the layout components
+             * @author Bedin Elia
+             */
+            @Override
+            public void onClick(View view) {
+                capturePhoto();
```
PrandiniUniPD, line 81: I'm sorry but maybe I wasn't explained well, I've meant that puts between 80 and 81 `buttonPhoto.setEnabled(false);` and removes the line 91-94. In this way the button when pressed once is not being pressed another time until the capturing of the photo is completed and so the capturePhoto method is not launched when the capturing of the photo is running. Could be a good idea?

ghost: In my opinion, it's clearer if the listener don't check anything, just calls the method that's supposed to do the action we are listening for.
And to be clear, I'm not saying your proposal is wrong, quite the opposite.
What do you think? Is it so critical that I should move the check inside the listener?

PrandiniUniPD: Rethinking about it with `buttonPhoto.setEnabled(false);` in the capturePhoto, the onClick of the listener and the capturePhoto are not launched until the `buttonPhoto.setEnabled(true);`, this behaviour is analogues to put that statement in the onClick of the listener, so it's not needed to put it in the listener, it works in both places. This is resolved!
I've written about removing also 91-92 because if the button is disabled the method capturePhoto is not launched actually. Could be `if (!buttonPhoto.isEnabled()) return;` of the line 92 useful now or in the future as a preventive checks or not in your opinion?

ghost: Actually, I forgot to take out the if statement from the old implementation, you are right.
It has the `!buttonPhoto.isEnabled()` because I did a _find and replace all_.
I'm going to take it out now thank you very much.

PrandiniUniPD (approved):

PrandiniUniPD:
The code review is completed, thanks to work with me. Have a nice day.

PrandiniUniPD:
Checklist:
- Comments about the code
- Explanations about the method or the code in general
- Modularity of the code
- Readability of the code
- Functionality relativeness to Android tools
- Respect of the Document Style proposed

PR #65, opened 2018-12-01
CodeReview of 16 from 2

ghost:
Here's the file for the review.
The introduction header should explain everything, but in case, fell free to PM me.
I wish you good work.

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
@@ -32,9 +32,8 @@ public class CameraActivity extends AppCompatActivity {
     //variable used to interface with the CameraView library
     private CameraView camera;
 
-    //variable (with default value false) set to true if a picture is captured
-    //only when the picture is sent to the PicturePreviewActivity the variable is set to false again
-    private boolean isPictureCaptured = false;
+    //variable used to connect to the capturePhoto button
+    private ImageButton buttonPhoto;
 
     /**
      * onCreate method of the Android Activity Lifecycle
@@ -47,7 +46,7 @@ public class CameraActivity extends AppCompatActivity {
         setContentView(R.layout.activity_camera);
         camera = findViewById(R.id.viewCamera);
 
-        //interface method for automatic onResume, onPause, onDestroy
+        //library method for automatic onResume, onPause, onDestroy
         camera.setLifecycleOwner(this);
 
         //Pinch to zoom support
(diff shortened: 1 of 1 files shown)


app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,123 @@
...
+        });
+
+    }
+
+    /**
+     * Method for taking a photo
+     * @author Bedin Elia
+     */
+    private void capturePhoto() {
+        if (mCapturingPicture) return;
```
leopardo-rossi-developer: Be careful because you've never initialize variable _mCapturingPicture_ so when the code is executed the first time it could throw an error you don't want. I suggest you to initialize it to _false_ in the onCreate method or directly after the declaration

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,123 @@
...
+import com.otaliastudios.cameraview.Gesture;
+import com.otaliastudios.cameraview.GestureAction;
+
+/**
+ * Activity for camera preview and taking photos
+ */
+public class CameraActivity extends AppCompatActivity {
+
+    private CameraView camera;
+    private boolean mCapturingPicture;
```
leopardo-rossi-developer: The name of this variable isn't very explanatory. If I were you, I'd call it something like _photoIsCaptured_

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,123 @@
...
+ *  The main goal for this project is to test older Camera API (15 to 19) to see if it's
+ *  plausible to support these devices and lower the minimum API requirement on CommonDemo
+ *  so we can support full range of Android devices (API 15 and higher is 100% demographic)
+ *  I used this interface for Camera (1) API to make the program
+ *      https://github.com/natario1/CameraView
+ *  working with master build of the interface
+ *  The documentation starts here
+ *      https://github.com/natario1/CameraView#usage
+ */
+
```
leopardo-rossi-developer, line 11: Very good explanation of this class' goal! In this way who reads this class can understand what is its purpose even before reading the whole code.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
mattia1142866: I agree with rossi.

mattia1142866 (commented):
the code is clear, the only uncertaintly is the inizialization of mCapturingPicture. I also notice that the code is commented in the right way.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: Thank you both very much. I agree I should have done that.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: Thanks for your feedback.
Yes, the name isn't as explanatory as it should

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
ghost: Thank you, very appreciated!

mattia1142866 (approved):
I read the corrections and the important things seem to be resolved

PR #66, opened 2018-12-01
code review of 6 from 13

TaulantBullaku:
Hello, can you review my code form 510 to 610

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
@@ -502,7 +502,7 @@ public class CameraActivity extends AppCompatActivity {
 
     /**
      *@param: communicate with the camera to have a picture
-     *@return: return the photo in a sharedPreferences
+     *@return: modifies shared preferences
      * Takes a photo.
      * <p>Saves the captured photo, previously converted into Base64 String, into the current
      * activity sharedPreferences</p>
@@ -566,7 +566,7 @@ public class CameraActivity extends AppCompatActivity {
             };
             mImageReader.setOnImageAvailableListener(readerListener, mBackgroundHandler);
 
-            //@author Leonardo Rosi
+            //@author Leonardo Rossi
 
             //Output surfaces
             List<Surface> outputSurface = new ArrayList<>(2);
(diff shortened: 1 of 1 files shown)


app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
```diff
@@ -0,0 +1,701 @@
...
+                        e.printStackTrace();
+                    }
+                }
+
+                @Override
+                public void onConfigureFailed(@NonNull CameraCaptureSession cameraCaptureSession)
+                {
+                }
+            },mBackgroundHandler);
+        } catch(CameraAccessException e) {
```
lucamoroz, line 613: Rather than enclosing all the code inside the try-catch block, you should insert inside the try-catch only the code that can throw an exception, so you can properly manage it.
You can check the doc:
https://developer.android.com/reference/android/hardware/camera2/CameraAccessException
too see which methods can throw a CameraAccessException and under what conditions.

app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
```diff
@@ -0,0 +1,701 @@
...
+                    byte[] photoByteArray = new byte[buffer.capacity()];
+                    buffer.get(photoByteArray);
+
+                    //Converts the byte array related to the given photo to a String in Base64 format
+                    String photoBitmapToString = Base64.encodeToString(photoByteArray, Base64.DEFAULT);
+
+                    //Create sharedPref file to save the Bitmap of last taken photo in a String form
+                    SharedPreferences sharedPref = getPreferences(Context.MODE_PRIVATE);
+                    SharedPreferences.Editor editor = sharedPref.edit();
+                    editor.putString("photoBitmap", photoBitmapToString);
```
lucamoroz: To avoid hard coding the key 'photoBitmap' you can save it using android String resources, it's very easy to use: https://developer.android.com/guide/topics/resources/string-resource#java

lucamoroz (commented):
There are just two little things to fix, for the rest the code seems fine to me!

app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
```diff
@@ -0,0 +1,701 @@
...
+        catch (CameraAccessException e)
+        {
+            e.printStackTrace();
+        }
+
+    }
+
+    /**
+     *@param: communicate with the camera to have a picture
+     *@return: return the photo in a sharedPreferences
```
lucamoroz: You should correct the specs: the method doesn't return a photo but modifies shared preferences.

lucamoroz (changes requested):
Found a specifications error.

PR #67, opened 2018-12-01
CodeReview of 6 from 19

TaulantBullaku:
Hi, can you review my code form 510 to 610?

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
@@ -559,14 +559,14 @@ public class CameraActivity extends AppCompatActivity {
                     String filePath= tempFileImage(CameraActivity.this, bmp,"capturedImage");
 
                     //An intent that will launch the activity that will analyse the photo
-                    Intent i = new Intent(CameraActivity.this, ResultActivity.class);
-                    i.putExtra("imageDataPath", filePath);
-                    startActivity(i);
+                    Intent intentActivity = new Intent(CameraActivity.this, ResultActivity.class);
+                    intentActivity.putExtra("imageDataPath", filePath);
+                    startActivity(intentActivity);
                 }
             };
             mImageReader.setOnImageAvailableListener(readerListener, mBackgroundHandler);
 
-            //@author Leonardo Rosi
+            //@author Leonardo Rossi
 
             //Output surfaces
             List<Surface> outputSurface = new ArrayList<>(2);
--- a/app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
@@ -559,9 +559,9 @@ public class CameraActivity extends AppCompatActivity {
                     String filePath= tempFileImage(CameraActivity.this, bmp,"capturedImage");
 
                     //An intent that will launch the activity that will analyse the photo
-                    Intent intentActivity = new Intent(CameraActivity.this, ResultActivity.class);
-                    intentActivity.putExtra("imageDataPath", filePath);
-                    startActivity(intentActivity);
+                    Intent intentCamera = new Intent(CameraActivity.this, ResultActivity.class);
+                    intentCamera.putExtra("imageDataPath", filePath);
+                    startActivity(intentCamera);
                 }
             };
             mImageReader.setOnImageAvailableListener(readerListener, mBackgroundHandler);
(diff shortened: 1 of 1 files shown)


app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
```diff
@@ -0,0 +1,701 @@
...
+                    editor.putString("photoBitmap", photoBitmapToString);
+                    editor.commit();
+
+                    //@author Leonardo Rossi
+                    //Temporary stores the caprured photo into a file that will be used from the Camera Result activity
+                    Bitmap bmp = BitmapFactory.decodeByteArray(photoByteArray, 0, photoByteArray.length);
+                    String filePath= tempFileImage(CameraActivity.this, bmp,"capturedImage");
+
+                    //An intent that will launch the activity that will analyse the photo
+                    Intent i = new Intent(CameraActivity.this, ResultActivity.class);
```
AndreaTonUnipd: You should avoid to name variables with too short and imprecise names. Abbreviations that are too short or, like in this case, just one letter can make your code less readable, even if the variable is used in two lines only and never again.

app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
```diff
@@ -0,0 +1,701 @@
...
+                    i.putExtra("imageDataPath", filePath);
+                    startActivity(i);
+                }
+            };
+            mImageReader.setOnImageAvailableListener(readerListener, mBackgroundHandler);
+
+            //@author Leonardo Rosi
```
AndreaTonUnipd: His surname is Rossi.
Still another question, should I review also the code written by someone else like Leonardo in this case ?

app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
```diff
@@ -0,0 +1,701 @@
...
+    /**
+     *@param: communicate with the camera to have a picture
+     *@return: return the photo in a sharedPreferences
+     * Takes a photo.
+     * <p>Saves the captured photo, previously converted into Base64 String, into the current
+     * activity sharedPreferences</p>
+     * @modify mCameraDevice
+     * @author Alberto Valente, Taulant Bullaku
+     */
+    private void takePhoto() {
```
AndreaTonUnipd, line 512: Wondering if this very big method could be divided in smaller ones to make it more modular. 
Also you could add more documentation about what you used and what you read to write this method, or maybe a tutorial.

app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
```diff
@@ -0,0 +1,701 @@
...
+
+    }
+
+    /**
+     *@param: communicate with the camera to have a picture
+     *@return: return the photo in a sharedPreferences
+     * Takes a photo.
+     * <p>Saves the captured photo, previously converted into Base64 String, into the current
+     * activity sharedPreferences</p>
+     * @modify mCameraDevice
```
AndreaTonUnipd: Since I have to review just a portion of this code, you should specify what mCameraDevice is, avoiding me to search across the whole class.

app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
```diff
@@ -0,0 +1,701 @@
...
+                }
+
+                @Override
+                public void onConfigureFailed(@NonNull CameraCaptureSession cameraCaptureSession)
+                {
+                }
+            },mBackgroundHandler);
+        } catch(CameraAccessException e) {
+            e.printStackTrace();
+        }
```
AndreaTonUnipd, line 611: I'd prefer more comments/documentation in the most complicated steps of the code, even just a small brief of the components you used.

AndreaTonUnipd (commented):
I found same example resembling this method but you did not link or suggest any documentation, anyway I don't get why you save the image as a string in sharedPreferences, is it useful to communicate with a database ? 
Also I think that shared Preferencies is an interface, not an Activity.

app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
TaulantBullaku: yes, because we wrote it together. 
I didn't write his last name. He did it.

app/src/main/java/unipd/se18/ocrcamera/cameraActiviti.java
TaulantBullaku: ok, thanks, I'll keep it in mind.

PR #68, opened 2018-12-01
Codereview of 11 from 18

BalzPit:
Review class BrightnessRecognition and line 45 and from 166 to 175 of CameraActivity

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/BrightnessRecognition.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/BrightnessRecognition.java
@@ -1,0 +1,93 @@
+package unipd.se18.ocrcamera;
+
+import android.content.Context;
+import android.graphics.Bitmap;
+import android.graphics.Color;
+import android.util.Log;
+import android.widget.Toast;
+import android.os.Handler;
+
+/**
+ * This class is used to calculate the brightness of an image in Bitmap format
+ * @author Pietro Balzan
+ */
+public class BrightnessRecognition {
+    Handler handler;             // handler for communication with the ui thread
+    private Context appContext;  //context of the activity that instances the object of this class
+    private final Bitmap image;  // the bitmap containing the image to analyze
+    private static final String TAG = "BrightnessRecognition";
+
+
+    /**
+     * Constructor of a BrightnessRecognition object containing a Bitmap
+     * @param bmp the bitmap of the image to calculate the brightness of
+     * @param cntxt context of the activity that instances an object of this class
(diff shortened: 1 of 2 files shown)


app/src/main/java/unipd/se18/ocrcamera/BrightnessRecognition.java
```diff
@@ -0,0 +1,93 @@
...
+        for(int x = 0; x < width; x+=pixelSkip) {
+            for (int y = 0; y < height; y+=pixelSkip) {
+                // get pixel color
+                pixel = image.getPixel(x, y);
+                R = Color.red(pixel);
+                G = Color.green(pixel);
+                B = Color.blue(pixel);
+
+                //  RGB/Luma conversion formula, dermines luminance of a pixel
+                double brightness = (0.2126 * R + 0.7152 * G + 0.0722 * B);
```
Fisher4537, line 63: Could you add a link to the documentation where to find the RGB/Luma conversion that you implement? I've search for some formula on the net but I didn't find the same formula. Is it a pragmatic formula?

Fisher4537:
I was thinking that we have many methods to edit images (rotateImage(), blurValue(), brightnessRecognition() and many other). I think that maybe it's time to implement an utility class for image manipulation, with static methods only. It would simplify the code and the testing. I asked the same question to @leopra, who implements blurValue() method. Maybe we could discuss together about it here! Let me know what do you think about it!

app/src/main/java/unipd/se18/ocrcamera/BrightnessRecognition.java
BalzPit: https://stackoverflow.com/questions/596216/formula-to-determine-brightness-of-rgb-color
I was browsing for a possible solution and found this, it seems to work quite well

BalzPit:
Yeah I think this is actually a very good Idea! we might have to talk about it with the rest of the class

Fisher4537:
If we agree, I think we should:
- close both the pull requests,
- open a new branch to with the new class, 
- collaborate to write the methods
- open a new pull request to merge with master

leopra:
I support the idea, now i'm closing the branch

app/src/main/java/unipd/se18/ocrcamera/BrightnessRecognition.java
Fisher4537: Thanks, now I can see which is the logic behind that formula! it would be nice if in the future code of the class for the manipulation of the images we had a comment inside the code with that link.

PR #69, opened 2018-12-01
Code review of 11 from 3

BalzPit:
Review class BrightnessRecognition and lines 45 and from 166 to 175 of CameraActivity

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/BrightnessRecognition.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/BrightnessRecognition.java
@@ -1,0 +1,93 @@
+package unipd.se18.ocrcamera;
+
+import android.content.Context;
+import android.graphics.Bitmap;
+import android.graphics.Color;
+import android.util.Log;
+import android.widget.Toast;
+import android.os.Handler;
+
+/**
+ * This class is used to calculate the brightness of an image in Bitmap format
+ * @author Pietro Balzan
+ */
+public class BrightnessRecognition {
+    Handler handler;             // handler for communication with the ui thread
+    private Context appContext;  //context of the activity that instances the object of this class
+    private final Bitmap image;  // the bitmap containing the image to analyze
+    private static final String TAG = "BrightnessRecognition";
+
+
+    /**
+     * Constructor of a BrightnessRecognition object containing a Bitmap
+     * @param bmp the bitmap of the image to calculate the brightness of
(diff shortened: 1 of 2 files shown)


app/src/main/java/unipd/se18/ocrcamera/BrightnessRecognition.java
```diff
@@ -0,0 +1,93 @@
...
+            handler.post(new Runnable() {
+                @Override
+                public void run() {
+                    Toast.makeText(appContext, tooDark, Toast.LENGTH_LONG).show();
+                }
+            });
+        }
+        else Log.d(TAG, "good image"); // image is neither too bright nor too dark
+    }
+}
```
gioppiva, line 93: This class is well written. The comments were good and allowed me to understand the code easily.
If i compare the code with the style document i can say that it's all okay.
I have just one question, what about if i have a Red, Green or Blue label? Will this method be successful?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,260 @@
...
+                    // analyze the brightness   @author Balzan Pietro
+                    final Bitmap bitmapImage2 = bitmapImage;
+                    new Thread(new Runnable() {      // run the brightness recognition on a new thread to improve performance
+                        public void run() {
+                            Context appContext= getApplicationContext();
+                            BrightnessRecognition bRecog = new BrightnessRecognition(bitmapImage2, appContext, handler);
+                            bRecog.imgBrightness(190,80,5);
+                        }
+                    }).start();
+
```
gioppiva, line 175: I can see no error in the code, the style document is respected and the comment are enough (also because i read the class building). I would like to know if this method can compromise the speed of execution. Overall it's ok.

BalzPit: When it wasn't on a separate thread the "takePhoto" method would wait for imgBrightness to finish its task before calling the intent to open the ResultActivity, so it did affect performance (there was a second where the program would just freeze) but running it on the new thread solved the problem

app/src/main/java/unipd/se18/ocrcamera/BrightnessRecognition.java
BalzPit: they are very rare cases but the method may fail on the red and blue case especially

PR #71, opened 2018-12-02
Codereview of 1 from 8

Nintenking:
The only class that I put and I ask you to review is "ImageProcessing", in which in collaboration with Thomas Porro from my group we are trying to auto-crop and auto-rotate the image. Every method should have written in their spec the authors so it should be easy for you to find which ones are made by me.
Thank you for your time!
EDIT: I accidentally committed some .ideas files, I have now deleted them


app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,349 @@
...
+
+
+    /**
+     * Converts the Bitmap into a grayscale matrix
+     * @param imagePath the matrix you want to convert
+     * @return the bitmap corresponding to the matrix
+     * @throws FileNotFoundException if the imagePath doesn't exist
+     * @author Oscar Garrido (g1)
+     */
+    public Mat conversion(String imagePath) throws FileNotFoundException {
```
PrandiniUniPD: There is a method with the same name but different sign at the line 249.
Is the behaviour similar to that method?
If not, could be a good Idea using a name more specific than only "conversion"? For example: `[...] Mat imgPathToGrayscaleMat(String imagePath) [...]`?
If yes, could be a good idea moving this method near the other method with the same name?

Could this method be private?

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,349 @@
...
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+    }
+
+
+    /**
+     * Converts the Bitmap into a grayscale matrix
+     * @param imagePath the matrix you want to convert
+     * @return the bitmap corresponding to the matrix
```
PrandiniUniPD: Is "the bitmap corresponding to the matrix" the same as "the Mat object generated by the processing of the Bitmap image"? Have I right understood?

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+
+    /**
+     * Converts a matrix into a Bitmap and saves it in a local directory
+     * @param matrix the matrix to be converted
+     * @param imageName the name of the file being saved
+     * @author Thomas Porro(g1), Oscar Garrido(g1)
+     */
+    private void save(Mat matrix, String imageName) {
```
PrandiniUniPD: Could be a good idea to return a String that will be null if the saving process will not have success and the String of the path if the pic will be correctly saved?

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,349 @@
...
+        final String directory = Environment.getExternalStorageDirectory() + "/" + Environment.DIRECTORY_PICTURES + "/ImageProcessingTest/" + imageName;
```
PrandiniUniPD: This method is private so I'm thinking about this is for an internal use of the class.
So could be a good Idea saving the pic in the cache of the app? Saving the pic in the cache avoids the problem of the writing permissions.

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,349 @@
...
+        Bitmap image = conversion(matrix);
+
+        OutputStream outStream = null;
+
+        File file = new File(directory);
+        file.mkdirs();
+
+        if (file.exists()) {
+            file.delete();
```
PrandiniUniPD: Could be dangerous deleting the file if it exists?

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,349 @@
...
+
+        //Crop the resulting image
+        getRectSubPix(rotatedImg, rect_size, rectangle.center, croppedImg);
+        return croppedImg;
+
+    }
+
+
+    /**
+     * Converts a matrix into a Bitmap and saves it in a local directory
```
PrandiniUniPD: What do you mean with "local"? Do you mean on the smartphone?

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,349 @@
...
+
+
+    /**
+     * Detect in which region of the picture there is some text
+     * @param imagePath the path of the image you want to analyze
+     * @return the rectangle which contains the text
+     * @throws FileNotFoundException if imagePath doesn't exist
+     * @author Thomas Porro (g1), Oscar Garrido (g1)
+     */
+    public RotatedRect detectMaxTextArea(String imagePath) throws FileNotFoundException {
```
PrandiniUniPD: Is this a method to find the text with the biggest area on the pic? Or does this method find a region with some text?

PrandiniUniPD (commented):
Hi, I've reviewed only the method that have your name in the javadoc comments, is it right?
I have some questions to you to completely understand your code, so please have patience.
I'm very interested to the library OpenCV but I've never used, so I'm happy learn some functionality about that.
Thanks in advance.

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: This method could very well be made private since it is only used for our own analysis!
The other "conversion" method is exactly the opposite of the one I wrote, I'll make sure that their name differ so that they will already describe themselves without confusion

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: This method is more for testing purposes because, in the end, we do not want to actually save in the gallery all the images created for our processes, but your idea is still good because we could avoid to hardcode the output directory and make the code cleaner!

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: As I said in a previous response, this method is used by us to check the output of the images we are creating in other methods using filter etc. so it's important that we could actually go to watch the image files in a specific directory to see if it is doing what we were expecting.
I don't think the method will remain implemented in the code by the end

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: No, it's exactly what we want to do since we don't want to save **all** the processed images, only the ones of the last photo we took!

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: Yes, you can see that the first line of the method is a directory **we** chose to *create* specifically *for this method* (it is hardcoded at the moment, so we will find a better way to do it)

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: If by "the Mat object generated by the processing of the Bitmap image" you mean the first Mat object that we are creating inside the method, then yes you understood correctly

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
PrandiniUniPD: For this reason I ask you if it's the case to save the file in the cache of the app. I'll share you a link that I think interesting about that https://developer.android.com/training/data-storage/files#WriteInternalStorage . This link was useful to me so I'll share you. I hope it is useful also to you.

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
PrandiniUniPD: Yep, maybe it is ok if you save in the internal cache of the app.
But if you use the standard path for the pictures and unfortunately another pic has the same name and it's not a pic generated by this app? For this reason I've asked you that.

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
PrandiniUniPD: I've seen it, thanks, but it's in the javadoc comments, usually the people see the javadoc and rarely the implementations of the code itself. Could be a good idea writing the specs more precisely for this reason?

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
PrandiniUniPD: Maybe there is a misunderstanding, this method return a Mat but in this line it's written that return a bitmap that corresponds to a matrix?

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: I'll make it clearer!

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: It's trying to find the (one) text region that has the biggest area

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: I'll try to implement the cache saving by tomorrow morning, thank you for the advice!

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: My fault, I wrote the `@return` wrong, I'll correct it in the next commit. Thank you

Nintenking:
I think I fixed all the issues you pointed out, could you check how is it now?

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
PrandiniUniPD: Decision to save in the cache, so, resolved

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
PrandiniUniPD: ok, thanks, could be a good idea splitting this method in more than one unique method? Anyway it's resolved.

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,348 @@
...
+     * @author Thomas Porro(g1), Oscar Garrido(g1)
+     */
+    private void save(Mat matrix, String imageName) {
+        
+        Bitmap image = conversionMatToBitmap(matrix);
+
+        OutputStream outStream = null;
+
+        //if not specified, the system-dependent default temporary-file directory will be used
+        File tmpFile = File.createTempFile(tmpPrefix);
```
PrandiniUniPD: Is this correct? Could be useful this https://developer.android.com/training/data-storage/files#WriteCacheFileInternal ?

PrandiniUniPD (commented):
I have only a question for you, good work!

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: Just to understand your question, you want to know if I should create a _Context_ object in which I would call the `.getCacheDir()` method, like in the example? If so, I was going to do it, but reading the Javadoc of `File.createTempFile` it should be unnecessary, since if the parameter of the directory is left blank the method will automatically pick the Cache Directory (I could better specify this in the above comment perhaps)

PrandiniUniPD: I'm sorry, I've not explained well, for the cache directory you are right but anyway is the suffix parameter needed too? I've seen it here https://developer.android.com/reference/java/io/File#createTempFile(java.lang.String,%20java.lang.String) .
And I don't find where the tmpPrefix is instantiated, have I missed something?
Thanks in advance.

Nintenking: Both my fault, I was so concentrated on the directory that I completely forgot the suffix. And you can't find where I initialized tmpPrefix because I didn't, sorry. I'll correct it now

PrandiniUniPD: Don't worry, never mind, it's important that we make good code together.

PrandiniUniPD (approved):

PrandiniUniPD:
The review is completed. Thanks for working with me. Have a good night.

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,349 @@
...
+        which require a large number of
+        parameters to be called, with a builder statement,
+        to simplify reading and eventually changing them
+        (at the moment there are a little too many "magic numbers")
+     */
+
+    //Tag used to identify the log
+    final String TAG = "openCV";
+
+
```
PrandiniUniPD, line 52: For the actual code style between methods (or in case similar to this) is permitted only one empty line, right?

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,349 @@
...
+    /**
+     * Detect in which region of the picture there is some text and finds the largest one
+     * @param imagePath the path of the image you want to analyze
+     * @return the rectangle which contains the text with maximum area
+     * @throws FileNotFoundException if imagePath doesn't exist
+     * @author Thomas Porro (g1), Oscar Garrido (g1)
+     */
+    public RotatedRect detectMaxTextArea(String imagePath) throws FileNotFoundException {
```
PrandiniUniPD: This is a method that made more things, could be useful to modularize it with some submethods?

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,349 @@
...
+
```
PrandiniUniPD: After the sign of the method is not permitted empty line, right? Have I remembered wrong?

PrandiniUniPD (commented):
In order to complete the review before Tuesday I've missed some things to check.
So I'm sorry it's my fault but I've some other things to speak about with you.
I hope to work right with you for learning and for making good code.

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: I'll correct it in every method, thank you

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: You're right, I was already working on it!

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: I'll correct it in every method

Nintenking:
I think I corrected all the possible changes you addressed, It should be fine now! I'm always available if you find anything else

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    /**
+     * Applies filters to an image do make it easier to detect rectangle areas
+     * @param imagePath the path of the image you want to analyze
+     * @return a matrix of the filtered image
+     * @throws FileNotFoundException if imagePath doesn't exist
+     * @author Thomas Porro (g1), Oscar Garrido (g1)
+     */
+	private Mat applyFilters(String imagePath) throws FileNotFoundException {
```
PrandiniUniPD: This is a private method but is it never used?

PrandiniUniPD: This is a private method but it is never used in this class. Is it right?

Nintenking: I just did the split of the method in 2 separate methods, to make more clear what exactly does each one of them. I haven't yet included these new methods in the rest of the code, I'll correct it later. Sorry for the inconvenience, I'll fix it asap

Nintenking:
I forgot to call the new "applyFilters" method after having split the original "detectMaxTextArea".  Now it should be fine, tell me what you think

PrandiniUniPD (approved):
I've done the review only of the code that Oscar Garrido has developed (not the Thomas Porro code) as requested.

Checklist:
- Comments about the code
- Explanations about the method or the code in general
- Modularity of the code
- Readability of the code
- Functionality relativeness to Android tools
- Respect of the Document Style proposed

The method ApllyFilters makes many things but we need some suggests to make it more modular and at the same time don't add weight to the code.

PrandiniUniPD:
The review is ended.
Thanks of working with me.
Have a nice day.

PR #73, opened 2018-12-02
Codereview of 1 from 14

Nintenking:
The only class that I put and I ask you to review is "ImageProcessing", in which in collaboration with Thomas Porro from my group we are trying to auto-crop and auto-rotate the image. Every method should have written in their spec the authors so it should be easy for you to find which ones are made by me.
Thank you for your time!
EDIT: I accidentally committed some .ideas files, I have now deleted them


app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,349 @@
...
+
+
+    /**
+     * Converts a matrix into a Bitmap and saves it in a local directory
+     * @param matrix the matrix to be converted
+     * @param imageName the name of the file being saved
+     * @author Thomas Porro(g1), Oscar Garrido(g1)
+     */
+    private void save(Mat matrix, String imageName) {
+        final String directory = Environment.getExternalStorageDirectory() + "/" + Environment.DIRECTORY_PICTURES + "/ImageProcessingTest/" + imageName;
```
frankplus: Hardcoding could be avoided by moving the path where the image is saved into a final variable of the class

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
```diff
@@ -0,0 +1,349 @@
...
+
+
+    /**
+     * Converts the Bitmap into a grayscale matrix
+     * @param imagePath the matrix you want to convert
+     * @return the bitmap corresponding to the matrix
+     * @throws FileNotFoundException if the imagePath doesn't exist
+     * @author Oscar Garrido (g1)
+     */
+    public Mat conversion(String imagePath) throws FileNotFoundException {
```
frankplus: There is an incorrect method overloading. There is Bitmap conversion(Mat matrix) and this one.
These two methods do different things one is converting a matrix into a Bitmap and this one is converting a matrix into a Bitmap. This can cause confusion. I suggest to rename them using more explicative names.

frankplus (commented):
In my opinion the code is sufficiently structured and documented. I haven't analyzed each image processing step because it goes beyond my expertise. A thorough test of the image processing effectiveness in improving the OCR confidence is needed. Good job for now

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: I didn't think about it, that way it should be more understandable compared to this "random" directory. Thank you!

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: I got the same comparison from my other reviewer and you're absolutely right; I'll make sure to change both the methods names so that they differ and they will describe themselves without confusion

app/src/main/java/unipd/se18/ocrcamera/ImageProcessing.java
Nintenking: In the end I completely changed the concept of the method: I was saving the file in an internal directory chosen by me, now I'm doing it into the system's default temp-file directory.

frankplus (approved):
good job I don't have anything else to say for now

PR #75, opened 2018-12-02
CodeReview of 12 from 4

NicoloCervo:
InciCsvCorrector.java and
PhotoTester at lines 379 to 403


InciCsvCorrector.java
```diff
@@ -0,0 +1,159 @@
+import java.io.*;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+/**
+ * Class for the correction of the csv file containing the INCI ingredients found at
```
thomasporro, line 6: I suggest to order the comments with the column limit characters of 100. So it will respect the javadoc style and it will be easier to read

InciCsvCorrector.java
```diff
@@ -0,0 +1,159 @@
...
+ * "http://ec.europa.eu/growth/tools-databases/cosing/pdf/COSING_Ingredients-Fragrance%20Inventory_v2.csv".
+ * There are reoccurring formatting errors in the downloaded file that make it unusable (line endings inside the ingredient description),
+ * this class fixes those errors by writing in a new file the fixed text, the paths to the source and end files will be given by the user.
+ * The INCI database is updated quite often so this class can be used every time there is an update to get a usable file.
+ * @author Nicolò Cervo (g3)
+ */
+public class InciCsvCorrector {
```
thomasporro, line 14: So if I have not misunderstood this class, you use it externally to the app. Am i right?

InciCsvCorrector.java
```diff
@@ -0,0 +1,159 @@
...
+        File corrected = new File(paths[1]);
+        boolean needsCorrection = false;
+        ArrayList<Integer> invLines = null;
+
+        try {
+            invLines = findInvalidLines(source);
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        if(!(invLines.isEmpty())) {   // if there are invalid lines print their number
```
thomasporro: You can delete the round brackets as you did below
```suggestion
        if(!invLines.isEmpty()) {   // if there are invalid lines print their number
```

InciCsvCorrector.java
```diff
@@ -0,0 +1,159 @@
...
+                e.printStackTrace();
+            }
+            System.out.println("the file has been corrected");
+        }else{
+            System.out.println("the file does not need correction");
+        }
+    }
+
+    /**
+     * finds the invalid lines in the source file and prints them
```
thomasporro, line 53: You should explain what an invalid line is

InciCsvCorrector.java
```diff
@@ -0,0 +1,159 @@
...
+
+        BufferedReader br = new BufferedReader(new FileReader(source));
+        String line;
+        int lineNum=0;
+        ArrayList<Integer> invalidLines = new ArrayList<>();
+
+        while (((line = br.readLine()) != null) ) {
+            lineNum++;
+            if(line.length()>MINIMUM_VALID_LENGTH) {
+                boolean currentLineHasValidStart = (isNumeric(line.substring(0, 5)) && (line.charAt(5) == ','));        //valid start of line (12345,)
```
thomasporro, line 68: You should explain the structure of a line. I can understand this line because i know that in the database there is a code for each ingredient

InciCsvCorrector.java
```diff
@@ -0,0 +1,159 @@
...
+        paths[0]= paths[0].replace("\\","\\\\"); // replace \ with \\ so the path can be used ("\" alone is an escape character)
+        System.out.print("Result file path :");
+        paths[1]= sc.nextLine();
+        paths[1]= paths[1].replace("\\","\\\\");
+
+        return paths;
+    }
+
+    /**
+     *
```
thomasporro: The "summary" of the method is missing

InciCsvCorrector.java
```diff
@@ -0,0 +1,159 @@
...
+    private static boolean isNumeric(String str)
+    {
+        for (char c : str.toCharArray())
+        {
+            if (!Character.isDigit(c)) return false;
+        }
+        return true;
+    }
+
+    /**
```
thomasporro, line 144: In this method you need to know if the date is valid or not? If not the method works, but if we have to know if the date is valid we need to update the method

InciCsvCorrector.java
```diff
@@ -0,0 +1,159 @@
...
+        BufferedWriter bw = new BufferedWriter(new FileWriter(corrected));
+        String line;
+        boolean currentLineHasValidStart;
+        boolean currentLineHasValidEnd;
+        boolean lineCompleted = true;
+
+        while((line = br.readLine())!=null){
+            if(line.length()>MINIMUM_VALID_LENGTH) {
+                currentLineHasValidStart = (isNumeric(line.substring(0, 5)) && (line.charAt(5) == ','));        //valid start of line (12345,)
+                currentLineHasValidEnd = isDate(line.substring(line.length() - DATE_LENGTH));              //valid end of line (dd/mm/yyyy)
```
thomasporro, line 98: Maybe you can create a variable that contains the substring, so in line 102 you don't have to do the substring method again

InciCsvCorrector.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+
+    /**
+     * finds the invalid lines in the source file and prints them
+     * @param source source file
+     * @return an ArrayList of integers with the line numbers in the source file
+     * @throws IOException when the file does not exist
+     */
+    private static ArrayList<Integer> findInvalidLines(File source) throws IOException {
```
thomasporro, line 58: You should throw a FileNotFoundException

InciCsvCorrector.java
```diff
@@ -0,0 +1,159 @@
...
+     * @return String[2] containing the paths of the source file [0] and the new corrected file [1]
+     */
+    private static String[] getFilesPaths(){
+
+        String[] paths = new String[2];
+        Scanner sc = new Scanner(System.in);
+
+        System.out.print("Source file path (C:\\Users\\...) :");
+        paths[0]= sc.nextLine();
+        paths[0]= paths[0].replace("\\","\\\\"); // replace \ with \\ so the path can be used ("\" alone is an escape character)
```
thomasporro: You don't have to replace the backslashes in the path, because if you pass the path, without modifying it, of the file to the File constructor, the object will be built without problems .
So this method isn't useful

InciCsvCorrector.java
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * gets the file paths from console input
+     * @return String[2] containing the paths of the source file [0] and the new corrected file [1]
+     */
+    private static String[] getFilesPaths(){
+
+        String[] paths = new String[2];
+        Scanner sc = new Scanner(System.in);
+
+        System.out.print("Source file path (C:\\Users\\...) :");
```
thomasporro: You might specify what kind of path I've to write  (absolute, relative, ...)

InciCsvCorrector.java
thomasporro: I've tried to do a test modifying the main and removing this class, and the program do all the stuff it have to do

InciCsvCorrector.java
```diff
@@ -0,0 +1,159 @@
+import java.io.*;
```
thomasporro: You should declare all the class you import so it's easier to read and in future if you add some lines it migth create some problems.

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    /**
+    * Returns a HashMap of (Tag, Value) pairs where value is the average test result of the photos tagged with that Tag
+    * @author Nicolò Cervo (g3) with the tutoring of Francesco Pham (g3)
+    */
+    public HashMap getTagsStats() throws JSONException {
+        HashMap<String, Float> tagStats = new HashMap<>(); //contains the cumulative score of every tag
+        HashMap<String, Integer> tagOccurrences = new HashMap<>(); //contains the number of occurrences of each tag
+        for(TestElement element : testElements) {
```
thomasporro, line 385: In the declaration of the variables you can add a comment to explain what testElements contains

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
```diff
@@ -0,0 +1,432 @@
...
+                    tagOccurrences.put(tag, tagOccurrences.get(tag) + 1);
+                }else{
+                    tagStats.put(tag, element.getConfidence());
+                    tagOccurrences.put(tag, 1);
+                }
+            }
+        }
+        Log.i(TAG, "getTagStats():");
+        for(String tag : tagStats.keySet()){
+            tagStats.put(tag, tagStats.get(tag)/tagOccurrences.get(tag)); // average of the scores
```
thomasporro, line 399: Maybe it's useful calculate the average of the score in an another line

InciCsvCorrector.java
```diff
@@ -0,0 +1,159 @@
...
+        String line;
+        boolean currentLineHasValidStart;
+        boolean currentLineHasValidEnd;
+        boolean lineCompleted = true;
+
+        while((line = br.readLine())!=null){
+            if(line.length()>MINIMUM_VALID_LENGTH) {
+                currentLineHasValidStart = (isNumeric(line.substring(0, 5)) && (line.charAt(5) == ','));        //valid start of line (12345,)
+                currentLineHasValidEnd = isDate(line.substring(line.length() - DATE_LENGTH));              //valid end of line (dd/mm/yyyy)
+                if (currentLineHasValidStart && currentLineHasValidEnd) {               //line is complete
```
thomasporro: I really don't understand the differences between the different bw.write(line). Can you explain it to me?

thomasporro (commented):
I've reviewed InciCsvCorrector.java and the method of photoTester.java that contains your name. Is it right?

InciCsvCorrector.java
NicoloCervo: yes, you just need to create one time the new file and then update the app to a new version. It didn't make much sense to put it in the app

InciCsvCorrector.java
NicoloCervo: Here I was not sure what to do because I already catch the file not found exception when i get the file so fileReader() should never throw an exception.

InciCsvCorrector.java
NicoloCervo: sure, will add an example

InciCsvCorrector.java
NicoloCervo: I commented the same lines in different ways, forgot to check the comments.
br.write(line + "\r\n");  writes the line end goes to the next while without "\r\n" it stays on the same line.

InciCsvCorrector.java
NicoloCervo: Ok, didn't know that.

InciCsvCorrector.java
NicoloCervo: Well, i tought of writing it in a more specific way but in the end i don't really care if the date is right or not (they should be self generated so it shouldn't be a problem) I just need to know if those 10 chars are in date format.
But i wouldn't hurt to check at least if the "/" are in the right positions, I'll fix that.

InciCsvCorrector.java
thomasporro: Perfect

InciCsvCorrector.java
thomasporro: 🔝

InciCsvCorrector.java
thomasporro: And this what kind of consequences have on the database?

NicoloCervo: it makes it so each line of the text file is an entry:
Ref No,INCI name,INN name,Ph. Eur. Name,CAS No,EC No,Chem/IUPAC Name / Description,Restriction,Function,Update Date

NicoloCervo: If a line ends with a date it should be complete so it gets written and the line ends, otherwise the next line of the source file must be the part that is missing from the current line so, without ending the line, I read the next and write it continuing on the same line.

NicoloCervo:
fixed

PR #78, opened 2018-12-02
CodeReview from 20 to 12 

RobertoUnipd18:


app/src/main/java/com/example/user/codereview/MainActivity.java
```diff
@@ -0,0 +1,114 @@
...
+
+    }
+
+    @Override
+    protected void onActivityResult(int requestCode, int resultCode, @Nullable Intent data) {
+
+        if (requestCode == REQUEST_IMAGE_CAPTURE && resultCode == RESULT_OK) {
+            button.setEnabled(true);
+            Bundle extras = data.getExtras();
+            imageBitmap = (Bitmap) extras.get("data");
```
NicoloCervo: maybe i'm missing something but shouldn't it be 
imageBitmap = (Bitmap) extras.get("BitmapImage");  ?

app/src/main/java/com/example/user/codereview/MainActivity.java
```diff
@@ -0,0 +1,114 @@
...
+import android.widget.EditText;
+import android.widget.ImageView;
+
+import java.net.URI;
+
+public class MainActivity extends AppCompatActivity implements View.OnClickListener {
+
+    private EditText text;
+    private ImageView photo;
+    private Button button;
```
NicoloCervo: It would probably make the code more readable if you named this button something related to it's function like you did with the other one

app/src/main/java/com/example/user/codereview/MainActivity.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+    }
+    
+
+    @Override
+    public void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
+        if(requestCode == 0){
+            if(grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED){
+                bTakePhoto.setEnabled(true);
```
NicoloCervo, line 64: can substitute method's body with
```
bTakePhoto.setEnabled( (grantResults.length > 0) && (requestCode == 0) && 
(grantResults[0] == PackageManager.PERMISSION_GRANTED));
```
but I'm not sure it makes the code more readable

app/src/main/java/com/example/user/codereview/MainActivity.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Override
+    public void onClick(View view){
+
+        switch(view.getId()) {
+
+            //this button change secondActivity
+            case (R.id.mainbutton):
+
+                String s = text.getText().toString();
```
NicoloCervo: Maybe use a more meaningful name fo this string

NicoloCervo (commented):
The code is overall easy to read, I only commented minor adjustments, it is good to use empty lines to separate logically blocks of statements but I found that some of them don't add anything to the code.

app/src/main/java/com/example/user/codereview/MainActivity.java
RobertoUnipd18: yes sure, i just change the name in bSwitchActivity

app/src/main/java/com/example/user/codereview/MainActivity.java
RobertoUnipd18: I've change it in sText

PR #80, opened 2018-12-02
CodeReview of 10 from 2

AlbVal97:
Hello Mattia, these are the files I would like you to review. The only fragments to consider are those authored by me or both me and other members: to be more specific there is the method takePhoto() and only a few lines of the class PostProcessingActivity. If you were asking, both of these are not currently in the most recent commonDemo version because they were updated or completely replaced. Let me know what do you think about it.

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
@@ -1,0 +1,761 @@
+package unipd.se18.ocrcamera;
+
+import android.Manifest;
+import android.app.Activity;
+import android.content.Context;
+import android.content.Intent;
+import android.content.SharedPreferences;
+import android.content.pm.PackageManager;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.graphics.ImageFormat;
+import android.graphics.Matrix;
+import android.graphics.SurfaceTexture;
+import android.hardware.Sensor;
+import android.hardware.SensorManager;
+import android.hardware.camera2.CameraAccessException;
+import android.hardware.camera2.CameraCaptureSession;
+import android.hardware.camera2.CameraCharacteristics;
+import android.hardware.camera2.CameraDevice;
+import android.hardware.camera2.CameraManager;
+import android.hardware.camera2.CameraMetadata;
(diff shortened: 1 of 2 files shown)


app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -0,0 +1,761 @@
...
+                @Override
+                public void onConfigureFailed(@NonNull CameraCaptureSession cameraCaptureSession)
+                {
+                }
+            },mBackgroundHandler);
+        } catch(CameraAccessException e) {
+            e.printStackTrace();
+        }
+    } //end takePhoto
+
```
mattia1142866, line 665: the code in this method is clean and clear and it has been commented well. I ask you why keep in the code the method onConfiguredFailed considering that is not used and has not been written.

app/src/main/java/unipd/se18/ocrcamera/PostProcessingActivity.java
```diff
@@ -0,0 +1,67 @@
...
+    public boolean onCreateOptionsMenu(Menu menu) {
+
+        MenuInflater inflater = getMenuInflater();
+        inflater.inflate(R.menu.crop_menu, menu);
+
+        //Menu reset button initialization
+        btnResetChanges = menu.findItem(R.id.btnResetChanges);
+
+        /**
+         * When clicked, resets the cropView to original rotation, zoom and crop frame
```
mattia1142866, line 54: very good explaination

mattia1142866 (approved):
that's a good code, I read good comments and the code is easy to read. i suggest that at the head of file write some lines of comments were explain the object of that code.

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
AlbVal97: Thank you Mattia! To be sincere I don't know why the method onConfiguredFailed has not been removed here, but it's very likely that in a more recent commit it would have been so. I need to recall that this file comes from a rather outdated commit due to the fact that the method takePhoto() has been completely refactored by now.

AlbVal97:
I think now it's ok!

PR #81, opened 2018-12-02
Codereview of 5 from 12

leopra:
Ciao questa è la mia parte di codice da revisionare, ho modificato takephoto e aggiunto i metodi blurValue e blurEvaluation

### DIFF
--- a/CameraActivity.java
+++ b/CameraActivity.java
@@ -1,0 +1,327 @@
+package unipd.se18.ocrcamera;
+
+import android.content.Context;
+import android.content.Intent;
+import android.content.SharedPreferences;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.graphics.Matrix;
+import android.hardware.Sensor;
+import android.hardware.SensorEvent;
+import android.hardware.SensorEventListener;
+import android.hardware.SensorManager;
+import android.os.Bundle;
+import android.support.design.widget.FloatingActionButton;
+import android.support.v7.app.AppCompatActivity;
+import android.util.Log;
+import android.view.View;
+import android.widget.Toast;
+
+import com.camerakit.CameraKitView;
+
+import org.opencv.android.OpenCVLoader;
+import org.opencv.android.Utils;
+import org.opencv.core.Core;
+import org.opencv.core.CvType;
+import org.opencv.core.Mat;
+import org.opencv.core.MatOfDouble;
+import org.opencv.imgproc.Imgproc;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.OutputStream;
+
+import static java.lang.String.valueOf;
+
+
+/**
+ * The Activity useful for making photos
+ */
+public class CameraActivity extends AppCompatActivity {
+
+    private CameraKitView cameraKitView;
+    private static String orientationResult;
+
+    /**
+     * onCreate method of the Android Activity Lifecycle
+     * @param savedInstanceState The Bundle of the last instance state saved
+     * @author Romanello Stefano
+     */
+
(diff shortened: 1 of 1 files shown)


CameraActivity.java
```diff
@@ -0,0 +1,336 @@
...
+    /**
+     * Detect the blurriness by appling a Laplacian matrix and evaluating the variance (low variance means an usually blurry image) uses OpenCV
+     * @param bitmap1 The captured image
+     * @return Double value of the variance
+     * @author Leonardo Pratesi
+     */
+    public static double blurValue(Bitmap bitmap1)
+    {
+        double blur;
+        double threshold= 10;
```
NicoloCervo, line 291: Instead of hardcoding 10 it  would be preferable to assign a it to a constant 
```
final BLUR_THRESHOLD = 10;
```

CameraActivity.java
```diff
@@ -0,0 +1,336 @@
...
+        Mat matImage = new Mat();
+        org.opencv.android.Utils.bitmapToMat(image, matImage);
+        Mat destination = new Mat();
+        Mat matGray=new Mat();
+
+        Imgproc.cvtColor(matImage, matGray, Imgproc.COLOR_BGR2GRAY);    //converte immagine in scala di grigi
+        Imgproc.Laplacian(matGray, destination, 3);              //applica prodotto di convoluzione a matGray e lo mette in destination
+        MatOfDouble median = new MatOfDouble();
+        MatOfDouble std= new MatOfDouble();
+        Core.meanStdDev(destination, median , std);                     //calcola deviazione standard
```
NicoloCervo, line 302: Not clear what median is used for, doesn't it stay empty?

CameraActivity.java
```diff
@@ -0,0 +1,336 @@
...
+
+                    SharedPreferences prefs = getSharedPreferences("prefs", MODE_PRIVATE);
+                    SharedPreferences.Editor edit = prefs.edit();
+                    edit.putString("imagePath", filePath.trim());
+                    edit.apply();
+
+                    //An intent that will launch the activity that will analyse the photo
+                    //goes to Result activity if the photo is not blurry [Leonardo Pratesi]
+                    if (!blurDetection(valoreBlur)) {
+                        Intent i = new Intent(CameraActivity.this, ResultActivity.class);
```
NicoloCervo, line 195:
```
Intent i = new Intent(CameraActivity.this, ResultActivity.class);
```

It would be better to avoid using single letters as names

CameraActivity.java
```diff
@@ -0,0 +1,336 @@
...
+         * AGGIUNGERE METODO CHE CALCOLA TRESHOLD IN BASE AL SOGGETTO FOTOGRAFATO
+         */
+        double threshold=10;
+
+        if (blurValue< threshold)
+            {
+                return true; //blurry
+            }
+        else
+                return false; //not blurry
```
NicoloCervo, line 329: You could subtitute this method's body with 
```
return(blurValue < BLUR_THRESHOLD);
```

NicoloCervo (commented):
I didn't find major issues, just proposed minor adjustments

CameraActivity.java
leopra: In reality that function is code already made by someone else

CameraActivity.java
leopra: Yes maybe, but i find it not very clear to understand

CameraActivity.java
leopra: The treshhold blur depends on the ambient of the photo so it is a parameter that has to be calculated everytime. Anyway I shouldn't have putted there thanks

PR #82, opened 2018-12-02
Codereview of 5 from 18

leopra:
Ciao questa è la mia parte di codice da revisionare, ho modificato takephoto e aggiunto i metodi blurValue e blurEvaluation

### DIFF
--- a/CameraActivity.java
+++ b/CameraActivity.java
@@ -1,0 +1,336 @@
+package unipd.se18.ocrcamera;
+
+import android.content.Context;
+import android.content.Intent;
+import android.content.SharedPreferences;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.graphics.Matrix;
+import android.hardware.Sensor;
+import android.hardware.SensorEvent;
+import android.hardware.SensorEventListener;
+import android.hardware.SensorManager;
+import android.os.Bundle;
+import android.support.design.widget.FloatingActionButton;
+import android.support.v7.app.AppCompatActivity;
+import android.util.Log;
+import android.view.View;
+import android.widget.Toast;
+
+import com.camerakit.CameraKitView;
+
+import org.opencv.android.OpenCVLoader;
+import org.opencv.android.Utils;
+import org.opencv.core.Core;
+import org.opencv.core.CvType;
+import org.opencv.core.Mat;
+import org.opencv.core.MatOfDouble;
+import org.opencv.imgproc.Imgproc;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.OutputStream;
+
+import static java.lang.String.valueOf;
+
+
+/**
+ * The Activity useful for making photos
+ */
+public class CameraActivity extends AppCompatActivity {
+
+    private CameraKitView cameraKitView;
+    private static String orientationResult="P";
+
+    /**
+     * onCreate method of the Android Activity Lifecycle
+     * @param savedInstanceState The Bundle of the last instance state saved
+     * @author Romanello Stefano
+     */
+
+
+    /**
+     * inizializza OpenCV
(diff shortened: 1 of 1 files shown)


Fisher4537:
Ciao Leonardo! Scusa se scrivo solo ora! Ho cominciato a leggere il codice, ma prima di cominciare volevo dirti che solitamente le review e i commenti vengono fatti in inglese, ma se preferisci possiamo farla in italiano! Fammi sapere cosa preferisci!

CameraActivity.java
```diff
@@ -0,0 +1,336 @@
...
+    public static Bitmap rotateImage(Bitmap source, int angle) {
+        Matrix matrix = new Matrix();
+        matrix.postRotate(angle);
+        return Bitmap.createBitmap(source, 0, 0, source.getWidth(), source.getHeight(),
+                matrix, true);
+    }
+
+    /**
+     * Detect the blurriness by appling a Laplacian matrix and evaluating the variance (low variance means an usually blurry image) uses OpenCV
+     * @param bitmap1 The captured image
```
Fisher4537, line 284: why don't we use a simple "bitmap" instead of "bitmap1"? which is the meaning of the "1"?

CameraActivity.java
```diff
@@ -0,0 +1,336 @@
...
+        double blur;
+        double threshold= 10;
+        Bitmap image = bitmap1;
+        Mat matImage = new Mat();
+        org.opencv.android.Utils.bitmapToMat(image, matImage);
+        Mat destination = new Mat();
+        Mat matGray=new Mat();
+
+        Imgproc.cvtColor(matImage, matGray, Imgproc.COLOR_BGR2GRAY);    //converte immagine in scala di grigi
+        Imgproc.Laplacian(matGray, destination, 3);              //applica prodotto di convoluzione a matGray e lo mette in destination
```
Fisher4537, line 299: Hardcode should be avoid! Why don't we define a local variable with a self-describing name?

CameraActivity.java
```diff
@@ -0,0 +1,336 @@
...
+        MatOfDouble median = new MatOfDouble();
+        MatOfDouble std= new MatOfDouble();
```
Fisher4537, line 301: Could I suggest a clearer name like "stdDev"? Or "standardDeviation"?

CameraActivity.java
```diff
@@ -0,0 +1,336 @@
...
+
+        return blur;
+    }
+
+    /** Detect the blurriness by appling a Laplacian matrix and evaluating the variance (low variance means an usually blurry image) uses OpenCV
+     * @param blurValue value from blurValue method
+     * @return boolean
+     * @author Leonardo Pratesi
+     */
+    public boolean blurDetection(double blurValue)
```
Fisher4537, line 316: Could I suggest a name that follow the java library like "isBlur"?

Fisher4537:
we haven't approved the style documentation yet therefore it's not necessary to follow the proposed documentation. However is strongly raccomanded to use only one style while coding.
You can find the proposed style documentation in #voci channel on Slack!

Fisher4537:
I was thinking that we have many methods to edit images (rotateImage(), blurValue(), blurDetection() and many other). We should implement a new class for image manipulation.
What do you think about it?

CameraActivity.java
leopra: Yes i will change

CameraActivity.java
leopra: Yes will be done thanks

CameraActivity.java
leopra: Yes i will keep it in mind

CameraActivity.java
leopra: Thanks for the observation

PR #83, opened 2018-12-02
CodeReview of 10 from 17

AlbVal97:
Hello Iosif, these are the files I would like you to review. The only fragments to consider are those authored by me or both me and other members: to be more specific there is the method takePhoto() into the class CameraActivity and only a few lines of the class PostProcessingActivity. I suggest to use Ctrl+F and look for "Alberto" to find them. If you were asking, both of these are not currently in the most recent commonDemo version because they were updated or completely replaced. Let me know what do you think about it.


app/src/main/java/unipd/se18/PostProcessingActivity.java
```diff
@@ -0,0 +1,67 @@
...
+        cropView = findViewById(R.id.cropView);
+        cropView.getCropImageView().setImageURI(captureImageUri);
+        cropView.getOverlayView().setShowCropFrame(true);
+        cropView.getOverlayView().setFreestyleCropMode(OverlayView.FREESTYLE_CROP_MODE_ENABLE);
+
+        //UI components initialization
+        btnConfirmCrop = findViewById(R.id.btnConfirmCrop);
+    }
+
+    @Override
```
MrVlado, line 44: Missing specifications for the method onCreateOptionsMenu

MrVlado (commented):
The method takePhoto looks good and pretty clear. The other method "btnResetChanges" it's clear enough  but you missed the completed specification.

app/src/main/java/unipd/se18/PostProcessingActivity.java
AlbVal97: This method has not been created by me, since it's clearly an overridden method of the class AppCompatActivity about the menu: https://developer.android.com/guide/topics/ui/menus

PR #87, opened 2018-12-06
Download images from DB

stefanoromanello:
Suggestions for the use:
Once you have downloaded everything try to download again. The algorithm skips all the already downloaded files, is a method to check if you have correctly downloaded all the images.


leopardo-rossi-developer:
I've done a push with an alternative solution with async task. In my opinion this solution improve yours because it separate the big method in smaller ones and then it handle better the communication with the UI thread. Take a look and tell me what do you think

stefanoromanello:
I thought about that but I didn't use the Async because I had to create another class. Anyway I will improve this because I don't like some parts and keep the async

leopardo-rossi-developer:
What don't you like?

leopardo-rossi-developer:
And yes you're right that it is another class, but in this way the code is more organized

stefanoromanello:
Fixed:
1) FTP Connection: was not set the remote directory
2) Buttons: as in the previous commit that Ive done, now the buttons disable itself preventing multiple downloads simultaneously and re-enable itself once finished
3) Most important: now the messages are live meaning you can see what the code is doing in the UI

leopardo-rossi-developer:
I've seen your updates and they're good! I've done a push with a little correction just to be more clear in the code

stefanoromanello:
Code cleaned, now should be finished.

PrandiniUniPD:
I'm sorry but today I haven't time to do the review but I've seen rapidly the code.
Could be useful make this activity between the result of the test activity but giving to the user the possibility of choosing the number of the photo to test (with the relative alterations)?
If the number of the photo is already on the smartphone -> ok, nothing to download;
otherwise -> download of that number of the photo and let's to analyse.

photo_2018-12-07_15-48-27

stefanoromanello:
for me no because there this is only for download the images and the other one for testing. If you want to implement what youre saying I suggest to code the testing activity to only read for example 10 image

PrandiniUniPD:
> for me no because there this is only for download the images and the other one for testing. If you want to implement what youre saying I suggest to code the testing activity to only read for example 10 image

But testing isn't possible without these pics downloaded, so anyway could be right unify the section of the testing with this download process?

app/src/main/java/unipd/se18/ocrcamera/PhotoDownloadTask.java
```diff
@@ -0,0 +1,216 @@
...
+    //private ArrayList<String> messages;
+    private Context context;
+    private ProgressBar progressBar;
+    private TextView textViewProgress;
+    private Button button;
+    private Integer currentProgress;
+    private ScrollView scrollCurrentDownload;
+    private TextView txtViewCurrentDownload;
+    //Constants
+    // PLEASE DON'T USE THESE FOR PERSONAL USAGE!
```
PrandiniUniPD: Could public the credentials here be danger? Providing an activity that asks the credentials could be more secure?

app/src/main/java/unipd/se18/ocrcamera/PhotoDownloadTask.java
```diff
@@ -0,0 +1,216 @@
...
+    // FTP CREDENTIALS
+    // USERNAME: epiz_22864730
+    // PASSWORD: 8M0tNtsJCsw
+    // HOSTNAME: ftpupload.net
+    // PLEASE DON'T USE THESE FOR PERSONAL USAGE!
+    // Communicate with Group 3 if you need something from the server.
+    private final String USERNAME = "epiz_22864730";
+    private final String PASSWORD = "8M0tNtsJCsw";
+    private final String HOSTNAME = "ftpupload.net";
+    private final String PHOTOS_FOLDER = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_PICTURES)+"/OCRCameraDB";
```
PrandiniUniPD, line 45: Could this variable be protected and static for using in the testing area?

PrandiniUniPD (commented):
It works on my smartphone (API level 27).
I'm in contact with @lucamoroz for a new design of the testing area because it doesn't work properly with a relatively large dataset with this dimension.
I'm not convinced about splitting the downloading of the test pics and the testing of these because this downloading is for doing tests. What do you think about it? Is really the case to split them?
Good use of the adapter, it makes the process efficient and usefully for us.
I've to do a more granular review but for now I would like to completely understand the work.
Thanks in advance.

app/src/main/java/unipd/se18/ocrcamera/PhotoDownloadTask.java
stefanoromanello: No, they're danger if you use them for other purposes.

app/src/main/java/unipd/se18/ocrcamera/PhotoDownloadTask.java
stefanoromanello: for now let them as they are please, I dont want the credential to be everywhere. I'm worried that someone use them in a bad way with ftpclient and risk to delete everything.

stefanoromanello:
For now lets keep them splitted but in "commond demo" so everyone can download the images. We can always do what you are saying, we will discuss next week

app/src/main/java/unipd/se18/ocrcamera/PhotoDownloadTask.java
PrandiniUniPD: I'm referring only to the line 49, I'm agree with you for the credentials.

app/src/main/java/unipd/se18/ocrcamera/PhotoDownloadTask.java
PrandiniUniPD: Could you restrict this user? Is it a good idea? This repo is public, is it dangerous having the credentials in the code or it could be better to ask to the user these credentials?

PrandiniUniPD:
> For now lets keep them splitted but in "commond demo" so everyone can download the images. We can always do what you are saying, we will discuss next week

Ok, this code is only for testing and it's urgent so we could accept to introduce this feature as is, but we should check some other things soon.

app/src/main/java/unipd/se18/ocrcamera/PhotoDownloadTask.java
stefanoromanello: PHOTOS_FOLDER declared in the same way is already used in other testing classes, no need to set this static

app/src/main/java/unipd/se18/ocrcamera/PhotoDownloadTask.java
stefanoromanello: I wanted to create other credentials but the free hosting doesn't let me do that

app/src/main/java/unipd/se18/ocrcamera/PhotoDownloadTask.java
PrandiniUniPD: Exactly, could be a good idea writing this variable once and sharing it to the other class that uses it?

app/src/main/java/unipd/se18/ocrcamera/PhotoDownloadTask.java
PrandiniUniPD: Ok, could be a good idea asking to the user these credentials instead writing it in this public code?

lucamoroz:
Right now the credentials are public, this should not happen because like this we make our database vulnerable. Please add a form to get the credentials from the user and change current password

stefanoromanello:
@lucamoroz this is only for testing. If we ask to the user (us) the credentials he has to put them manually and he knows anyway the credentials. Is not possible to change the password of the ftp

lucamoroz:
This repository is public and  now anyone can read the credentials and potentially delete our data, we can't leave this possibility

stefanoromanello:
Updated. Now java reads the credentials from a file located in the Document directory. I will change the password manually and upload in slack the file with the credentials

PrandiniUniPD (approved):
This is an urgent feature useful for testing.
So we could introduce it now.
Eventually we could improving or adjusting it during the using of it.

PR #88, opened 2018-12-08
Testing issues fix

lucamoroz:
Solved memory leak caused by loading all the photos on the heap memory (now class TestElement stores the image path rather than the full image), covered NaN case in function ingredientsTextComparison


frankplus (approved):

PrandiniUniPD:
Outdated

PR #89, opened 2018-12-08
Implemented showing ingredients informations from INCI database 

frankplus:
I implemented the search of ingredients extracted from the OCR inside the INCI database. I used the weighted levenshtein algorithm to search for the most similar ingredient for each comma separated text taken from the OCR. The result is shown in a list view sorted by best similarity. 
Suggestions are welcomed.


GiovanniFurlan:
hi, I'm working in inci research too. I see ur solution, it's good in theory but it needs too much time to process the text, for a text of 20 lines it takes at least 20/30 seconds.  The the match could be very very strange so could be good to not show elements with similarity under a good %. The app find "SD ALCOHOL 23-A"  from the string "Learn more".
The solution i think and implement is to search from the database to the text, but the database is bad written so it's hard to find a good delimiter, "/"  has 2 different uses so it can't be used.

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,133 @@
...
+     */
+    private Ingredient findBestMatchingIngredient(String ingredient){
+        //ignoring case by converting to upper case like all texts in database
+        ingredient = ingredient.toUpperCase();
+
+        LevenshteinStringComparator stringComparator = new LevenshteinStringComparator();
+        double maxSimilarity = -1;
+        int bestMatchingIngredient = -1;
+        for(int i = 0; i< listIngredients.size(); i++){
+            double similarity = stringComparator.getNormalizedSimilarity(listIngredients.get(i).getInciName(), ingredient);
```
lucamoroz, line 89: Is there a way to avoid calculating the exact similarity for each ingredient? for example, could we stop the evaluation before if we find out that we can't do better than 'similarityThreshold'? Maybe this could improve the performance

app/src/main/java/unipd/se18/ocrcamera/IngredientsActivity.java
```diff
@@ -0,0 +1,81 @@
...
+        SharedPreferences prefs = getSharedPreferences("prefs", MODE_PRIVATE);
+        String OCRText = prefs.getString("text", null);
+
+        if(OCRText != null) {
+            InciThread inciThread = new InciThread(OCRText);
+            inciThread.start();
+        }
+    }
+
+    class InciThread extends Thread{
```
lucamoroz, line 37: Why are you using a thread here? couldn't you run this piece of code directly on the main thread?

app/src/main/java/unipd/se18/ocrcamera/LevenshteinStringComparator.java
```diff
@@ -0,0 +1,60 @@
...
+ * @author Francesco Pham
+ */
+public class LevenshteinStringComparator extends WeightedLevenshtein{
+
+    public LevenshteinStringComparator(){
+        super(
+            new CharacterSubstitutionInterface() {
+                public double cost(char c1, char c2) {
+
+                    // The cost for substituting 't' and 'r' is considered
```
lucamoroz: Why are you reducing the cost of substituting these chars even if we are not using a keyboard? Couldn't these params all be '1'?
As we are using an OCR, we could reduce the cost of substituting similar chars (e.g. T and I, D and O ...), what do you think?

lucamoroz (commented):
I asked some questions, anyway I think we should wait before pushing this into CommonDemo as we have to decide the design of the implementation

app/src/main/java/unipd/se18/ocrcamera/Inci.java
frankplus: I thinks it is better if we find the best matching ingredient and not the first ingredient that reach the threshold. We won't improve the performance much better doing that.

app/src/main/java/unipd/se18/ocrcamera/IngredientsActivity.java
frankplus: Because the processing takes a long time we can't stop the main UI thread, it would crash the app.

app/src/main/java/unipd/se18/ocrcamera/LevenshteinStringComparator.java
frankplus: Sorry that comment was copied from elsewhere. Yes you are right, we could change and add more similar characters to the method.

app/src/main/java/unipd/se18/ocrcamera/IngredientsActivity.java
lucamoroz: I mean, if you are comparing 'aqua' with 'gas' (two words that have nothing in common) can we stop the evaluation of the levenshtein distance before (for example noticing that the distance is greater than 3) and save some time? This reasoning could be done for many other ingredients

app/src/main/java/unipd/se18/ocrcamera/Inci.java
lucamoroz: I mean, if you are comparing 'aqua' with 'gas' (two words that have nothing in common) can we stop the evaluation of the levenshtein distance before (for example noticing that the distance is greater than 3) and save some time? This reasoning could be done for many other ingredients

frankplus:
This implementation compares each block of text separated by commas with each ingredient's INCI name. This method of extracting ingredients from the OCR text takes from 10 to 20 seconds which has to be improved. The reason why it takes this time is because of the levenshtein quadratic complexity  which is O(n*m) where n and m are the length of string compared. There is also a different problem: this method can't find ingredients which are not properly separated by commas. 
I'm trying to find a different way of solving the search all the ingredients inside the text and more  efficiently. I'm closing this pull request but leaving the branch as it is, continuing on another branch. Thank you for your time.

PR #93, opened 2018-12-14
Extraction and display ingredients informations

frankplus:
This is somehow a continuation of pull request #89 
Changelog:
- Load INCI database from csv file into a data structure
- Ingredients extraction: I actually implemented two different extraction methods, both very efficient.
    - _TextSplitIngredientsExtractor_: this is the same algorithm used in pull request #89. It splits the ocr text and for each block of text search inside INCI DB for the ingredient with minimum distance using Levenshtein. It is a lot more efficient now because using bk-trees, the complexity is logarithmic instead of quadratic. However the **downside** of this method is that ingredients that are not properly separated are not recognized. I'm still using this for testing purposes.
    - _PrecorrectionIngredientsExtractor_: this class tries to correct the text before using a dictionary of different words taken from INCI DB (the dictionary is much smaller than entire DB). Then for each INCI ingredient check if it is contained inside the corrected text. This is what I plan to use for the extraction because it doesn't have the cons of the former method. The text correction could be improved even more managing cases such as line break etc..
- Ingredients display on a list view in result activity also showing progress dialogs.
- Added tests for these extractions inside PhotoTester.
- Added JUnit tests, testing text correction and ingredients extraction.
- This also include changes made in photo_testing branch by Moroldo and Prandini.

As always, suggestions are welcomed.


frankplus:
I just added some other text correction managing line breaks:
- If there is a hyphen preceding a line break, then merge current line and next line together 
(e.g. "COPOL-
YMER" becomes "COPOLYMER").
- If there are line breaks (without hyphen) then substitute line break with a whitespace

PR #94, opened 2018-12-15
Manual Cropping (This is a pointer)

PrandiniUniPD:
This is a pointer to the manual method for cropping the image.
It's the method proposed at the last deadline.
I've restored it for the people that have the necessity of this method but it's only a pointer so please don't merge this pull request if it's not useful, other groups are working for developing an automatic method to do this (for this reason I've only restored it without any preparation for code reviewing).


PrandiniUniPD:
I've opened this pull request for making a pointer about the manual cropping viewable from all the people.
For cleaning the open pull request list I will close this one but in the closed pull request list will remain this pointer for the people that could need it.

PR #95, opened 2018-12-15
Gallery Activity + Improvement for download activity

stefanoromanello:
This is a gallery for the taken images (not for the testing images).
It uses the meta-tags of the image for storing all the important informations and fragments for managing the main gallery and the detail view page.

For this code I needed the last downloadDbActivity Commit so this pull request includes also the last version of the download Activity which is already in the common demo but this new version adds some few permission verifications


Nintenking:
Both the automatic login (with the .txt file in Documents) and the manual login work just fine; the gallery shows exactly all and only the last taken images via our app.  
This pull request is fine by me!

Nintenking (approved):

PrandiniUniPD:
Outdated

PR #96, opened 2018-12-16
Avoiding the single point of failure (Text extraction as suggested by the Doctor Li ) + Demo presented at the last deadline

PrandiniUniPD:
This pull request aims to avoid the single point of failure by using the interface properly and a specific package for the extraction.
The extraction with ml-kit works with a listener so the thread is not stopped and so I've measured an increasing of the performance of the recognizing process.
Also it's included some improvement to the test section for viewing the result of the tests, some of these are:
- view of the element cleaner than before
- specific button to see the details of every test
- progress bar added

**_Edit_**
See https://github.com/PrandiniUniPD/OCRCamera/pull/96#issuecomment-448646869


Fisher4537:
Why don't we divide the purposes of this branch on different branches? Each branch should implements only one feature. So if a branch is approved before the other, we could merge it with CommonDemo and everybody who is working with CommonDemo can pull out the new feature before!

lucamoroz:
I totally approve the renovation of the OCR, but I found that the changes on PhotoTester makes that it's harder to understand how a single test works, as we need to jump through many methods. I would have just blocked the thread calling getTextFromImg() with a semaphore.

PrandiniUniPD:
> Why don't we divide the purposes of this branch on different branches? Each branch should implements only one feature. So if a branch is approved before the other, we could merge it with CommonDemo and everybody who is working with CommonDemo can pull out the new feature before!

If you are referring to the viewing of the tests result, I had to include that because it's not working with this new implementation of the text extraction because it's true that there was an interface to avoid a single point of failure but it wasn't used properly.
Changing that as suggested I had to modify also the testing section.

PrandiniUniPD:
Later than the comment of @Fisher4537 I've pushed some commits about the work of @frankplus with the scope of presenting the work at the deadline for tomorrow, so I'm closing this pull request and I will open a new pull request with only the work relative to the extraction of the text (but obviously it will include changing to the testing section also).

PrandiniUniPD:
### Change of mind
#### The Fact
This branch has become a branch very similar to CommonDemo because in this one there is the merging of the CommonDemo branch, #88, #93, #95, the code that starts this pull request #96 and some bugfix introduced later than the various merging and these bugfix are only in this branch.
#### The Context
We need all the features presented and we need many time to improve the our new CommonDemo for the next deadline.
We have to consider that the CommonDemo branch is not the "master", it's firstly used for testing and have a code to base on for new features or adjusting them, when the CommonDemo will contain a very good code, it will be introduced to the master branch.
#### The Idea
Splitting the features and the bugfix introduced here could be hard and it couldn't be so useful than having a new CommonDemo to base on for starting the work for the next sprint.
We can review the singular part anyway.
For example every people can refer to the singular pull request to view only the changes proposed by that and I'll add the eventual updates here or every people can review a singular piece of this pull request.
So I'll reopen this pull request.

PrandiniUniPD:
Outdated

PR #98, opened 2018-12-17
Brightness Recognition

BalzPit:
This branch contains the "BrightnessRecognition" class, used to determine wether an image is under or overexposed. Its method "imgBrightness" is called on a new thread in ResultActivity and prompts a message to the user via toast if the image is not optimal.


PrandiniUniPD:
Outdated

PR #99, opened 2019-01-02
Demo g2 + g3

frankplus:
This demo is a merge of:

- pull request #96 from restructuring:

> This pull request aims to avoid the single point of failure by using the interface properly and a specific package for the extraction.
The extraction with ml-kit works with a listener so the thread is not stopped and so I've measured an increasing of the performance of the recognizing process.
Also it's included some improvement to the test section for viewing the result of the tests, some of these are:
view of the element cleaner than before
specific button to see the details of every test
progress bar added

- pull request #93 from textAutoCorrection: which comprises the following implementations 
    - **Inci**: Load INCI database from csv file into a data structure. 
    - **TextAutoCorrection**: automatic correction of the ocr text using a dictionary of different words taken from INCI DB.
    - **IngredientsExtractor**: an extractor of a list of ingredients from the ocr text
- pull request #98 from Brightness_recognition: 

> This branch contains the "BrightnessRecognition" class, used to determine wether an image is under or overexposed. Its method "imgBrightness" is called in an AsyncTask in ResultActivity and prompts a message to the user via toast if the image is not optimal.
- pull request #95 from galleryActivity: 

> This is a gallery for the taken images (not for the testing images).
It uses the meta-tags of the image for storing all the important informations and fragments for managing the main gallery and the detail view page.
For this code I needed the last downloadDbActivity Commit so this pull request includes also the last version of the download Activity which is already in the common demo but this new version adds some few permission verifications

- New feature: when clicked on an ingredient pops up a fragment showing its informations including a request to wikipedia.

- New feature: manual search of an ingredient and show informations about the ingredient selected.


PrandiniUniPD:
Outdated

PR #101, opened 2019-01-10
CodeReview #2 of 14 (Romanello) from 7 (Prandini)

stefanoromanello:
The code review is for the part of the app regarding the Gallery Activity

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
@@ -1,0 +1,385 @@
+package unipd.se18.ocrcamera;
+
+import android.Manifest;
+import android.content.DialogInterface;
+import android.content.pm.PackageManager;
+import android.os.Build;
+import android.support.v4.app.ActivityCompat;
+import android.support.v4.app.Fragment;
+import android.support.v4.app.FragmentManager;
+import android.support.v4.app.FragmentTransaction;
+import android.support.v4.content.ContextCompat;
+import android.support.v7.app.ActionBar;
+import android.support.v7.app.AlertDialog;
+import android.support.v7.app.AppCompatActivity;
+import android.os.Bundle;
+import android.support.v7.widget.DefaultItemAnimator;
+import android.support.v7.widget.GridLayoutManager;
+import android.support.v7.widget.RecyclerView;
+import android.view.LayoutInflater;
+import android.view.Menu;
+import android.view.MenuInflater;
+import android.view.MenuItem;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.ImageView;
+import android.widget.TextView;
+import java.io.IOException;
+import java.util.ArrayList;
+
+
+/**
+ * Gallery activity
+ * @author Stefano Romanello
+ */
+public class GalleryActivity extends AppCompatActivity {
+
+    //Code for internet permission
+    private static final int REQUEST_PERMISSION_CODE = 500;
+    private FragmentManager fm;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState)
+    {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_gallery);
+        fm = getSupportFragmentManager();
+
+        //When the activity starts the first thing that I do i check for storage permission
+        //If the function returns true I'll load the home page.
+        if(verifyStoragePermission())
+        {
+            loadHomeFragment();
+        }
+    }
+
+    /**
+     * Function for load the home fragment from onActivityCreated and
+     * onRequestPermissionsResult in case I get the storage permission from the permission
+     */
+    private void loadHomeFragment()
+    {
+        FragmentTransaction ft = fm.beginTransaction();
+        ft.add(R.id.fragmentPlaceHolder, new MainFragment(),getString(R.string.homeFragmentTag));
+        ft.commit();
+    }
+
+    @Override
+    public void onBackPressed()
+    {
+        int backStackEntryCount = getSupportFragmentManager().getBackStackEntryCount();
+        //If there is more than 1 fragment running I close the
+        //current one (DetailsFragment) and return to the prevous (HomeFragment)
+        //Or I just call the default onBackPressed which closes the activity
+        if (backStackEntryCount != 0) {
+            fm.popBackStack();
+        } else {
+            super.onBackPressed();
+        }
+    }
+
+
+
+    /************************************************************/
+    /***************************FRAGMENTS************************/
+    /************************************************************/
+
+
+    /**
+     * Fragment of the gallery layout
+     * @author Romanello Stefano
+     */
+    public static class MainFragment extends Fragment
+    {
+        GalleryManager.RecycleCardsAdapter cardAdapter;
+        /**
+         * Event triggered once everything in the activity have finished loading
+         */
+        @Override
+        public void onActivityCreated(Bundle savedInstanceState)
+        {
+            super.onActivityCreated(savedInstanceState);
+            ActionBar actionBar =((GalleryActivity)getActivity()).getSupportActionBar();
+            actionBar.setTitle(R.string.galleryFragmentTitle);
+            actionBar.setDisplayHomeAsUpEnabled(false);
+            actionBar.setHomeButtonEnabled(false);
+        }
+
+        /**
+         * Event triggered when I create the view
+         */
+        @Override
+        public View onCreateView(LayoutInflater inflater, ViewGroup parent, Bundle savedInstanceState)
+        {
+            // Defines the xml file for the fragment
+            return inflater.inflate(R.layout.activity_gallery_content, parent, false);
+        }
+
+        /**
+         * This event is triggered soon after onCreateView().
+         * Any view setup should occur here.
+         */
+        @Override
+        public void onViewCreated(View view, Bundle savedInstanceState)
+        {
+            //Create the container for all my cards
+            RecyclerView picturesRecycleView = view.findViewById(R.id.recycle_view);
+            picturesRecycleView.setHasFixedSize(true);
+
+            //Set a grid (2 columns) from my recycleView
+            GridLayoutManager recycleLayoutManager = new GridLayoutManager(view.getContext(), 2, GridLayoutManager.VERTICAL, false);
+            picturesRecycleView.setLayoutManager(recycleLayoutManager);
+            picturesRecycleView.setItemAnimator(new DefaultItemAnimator());
+
+            //Load the cards using the RecycleCardsAdapter into the picturesRecycleView
+            ArrayList<GalleryManager.PhotoStructure> photos = GalleryManager.getImages(view.getContext());
+            cardAdapter = new GalleryManager.RecycleCardsAdapter(view.getContext(), photos);
+            picturesRecycleView.setAdapter(cardAdapter);
+        }
+
+
+        /**
+         * Event when the main fragment change its hidden status.
+         * @param hidden true=is hidden, false=is not hidden.
+         */
+        @Override
+        public void onHiddenChanged(boolean hidden)
+        {
+            //I have to understand if the user has deleted an image
+            //If the answer is yes I have to remove the item from the recycleView
+            if(!hidden)
+            {
+                //When I delete something from the DetailsFragment using the method GalleryManager.deleteImage(photoInfos)
+                //I set in the GalleryManager an object containing the PhotoStructure to delete.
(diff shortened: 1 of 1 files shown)


app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
```diff
@@ -0,0 +1,385 @@
...
+import android.widget.ImageView;
+import android.widget.TextView;
+import java.io.IOException;
+import java.util.ArrayList;
+
+
+/**
+ * Gallery activity
+ * @author Stefano Romanello
+ */
```
PrandiniUniPD, line 34: I know the work you are working into and so I understand it, but could a good idea be writing more of this in the description of this class for extending the range of the people that can understand it?

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
```diff
@@ -0,0 +1,385 @@
...
+public class GalleryActivity extends AppCompatActivity {
+
+    //Code for internet permission
+    private static final int REQUEST_PERMISSION_CODE = 500;
```
PrandiniUniPD, line 38: In this class, this permission code seems to be used for request the writing storage permission and not an "internet permission". Could a good idea be making the name of this permission code more specific in order to make a more clear scope for this variable?

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
```diff
@@ -0,0 +1,385 @@
...
+    /************************************************************/
+    /***************************FRAGMENTS************************/
+    /************************************************************/
+
+
+    /**
+     * Fragment of the gallery layout
+     * @author Romanello Stefano
+     */
+    public static class MainFragment extends Fragment
```
PrandiniUniPD, line 92: This class is public so another activity could use it without launching the check of the permissions required.
In the other hand this class is useful as it is (public) because the design of the UI could be change in the future and the activity (or activities) that hosts this fragment could be different.
So could a good idea including in the MainFragment class the check about the permissions that it needs and splitting this class (GalleryActivity) in two classes, one only for the activity and one only for the fragment?

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
```diff
@@ -0,0 +1,385 @@
...
+    {
+        GalleryManager.RecycleCardsAdapter cardAdapter;
+        /**
+         * Event triggered once everything in the activity have finished loading
+         */
```
PrandiniUniPD, line 97: There is used a javadoc notation for this comment and for others next comments.
Would you like to used it or it would be only a comment?
If you would like to make a javadoc comment the "@param savedInstanceState" is missing, but if you would like to make only a comment I suggest you to consider to make the notation used for only the normal comments.

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
```diff
@@ -0,0 +1,385 @@
...
+    /************************************************************/
+
+
+    /**
+     * Fragment of the gallery layout
+     * @author Romanello Stefano
+     */
+    public static class MainFragment extends Fragment
+    {
+        GalleryManager.RecycleCardsAdapter cardAdapter;
```
PrandiniUniPD, line 94: Could a good idea be putting a description about GalleryManager or the GalleryManager.RecycleCardAdapter?
I have an idea about that (an adapter the prepare CardViews that will be hosted by a RecyclerView?) but from the code it's not clear.

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
```diff
@@ -0,0 +1,385 @@
...
+        /**
+         * Event triggered once everything in the activity have finished loading
+         */
+        @Override
+        public void onActivityCreated(Bundle savedInstanceState)
+        {
+            super.onActivityCreated(savedInstanceState);
+            ActionBar actionBar =((GalleryActivity)getActivity()).getSupportActionBar();
```
PrandiniUniPD, line 102: Is this statement specific for the use with the GalleryActivity only? If yes, could a good idea be doing this and the next statements about the action bar in the GalleryActivity and not here for supporting modularity and usability?

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
```diff
@@ -0,0 +1,385 @@
...
+        public View onCreateView(LayoutInflater inflater, ViewGroup parent, Bundle savedInstanceState)
+        {
+            // Defines the xml file for the fragment
+            return inflater.inflate(R.layout.activity_gallery_content, parent, false);
+        }
+
+        /**
+         * This event is triggered soon after onCreateView().
+         * Any view setup should occur here.
+         */
```
PrandiniUniPD, line 121: Could more useful than this be describing what things do this method? When it is launched is in the Android documentations and it is possible putting a link about that.

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
```diff
@@ -0,0 +1,385 @@
...
+            ArrayList<GalleryManager.PhotoStructure> photos = GalleryManager.getImages(view.getContext());
+            cardAdapter = new GalleryManager.RecycleCardsAdapter(view.getContext(), photos);
+            picturesRecycleView.setAdapter(cardAdapter);
+        }
+
+
+        /**
+         * Event when the main fragment change its hidden status.
+         * @param hidden true=is hidden, false=is not hidden.
+         */
```
PrandiniUniPD, line 144: With "hidden status" the description is general, and a coder rarely see the implementation of the code in order to understand the things that a method does.
Could a good idea be describing the method with a non-general description?

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
```diff
@@ -0,0 +1,385 @@
...
+
+                int deletedPosition=cardAdapter.getDeletedPhotoPosition();
+                if(deletedPosition!=-1)
+                {
+                    //have to remove the photo from the recycler
+                    cardAdapter.photosList.remove(deletedPosition);
+                    cardAdapter.notifyItemRemoved(deletedPosition);
+                }
+
+                //Restore the actionBar
```
PrandiniUniPD, line 165: The lines of code from the line 166 to the line 169, the lines of code from the line 191 to the line 194 and the lines of code from the line 102 to the line 105 are the same.
If they don't specialize this fragment only for the use with the GalleryActivity, could a good idea be putting them to a single method that is launched every times it's needed?

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
```diff
@@ -0,0 +1,385 @@
...
+                actionBar.setHomeButtonEnabled(false);
+            }
+        }
+    }
+
+    /**
+     * Fragment of the detailed photo layout
+     * @author Romanello Stefano
+     */
+    public static class DetailFragment extends Fragment
```
PrandiniUniPD, line 178: Which class use this class named DetailFragment? It seems that it's not launched by GalleryActivity or GalleryActivity.MainFragment. Is it right?
If yes, could a good idea be making a single class for this part of code to improve modularity and to make the code clearer than this?

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
```diff
@@ -0,0 +1,385 @@
...
+            }
+        }
+    }
+
+    /**
+     * Verify if the app has the storage permission
+     * @return boolean of the current status (before asking the permission)
+     * if its false this prevent the app to load the fragment that uses the internal storage
+     */
+    public boolean verifyStoragePermission()
```
PrandiniUniPD, line 346: The "Manifest.permission.WRITE_EXTERNAL_STORAGE" is for having the permission for deletion, right?
Does it need also the "Manifest.permission.READ_EXTERNAL_STORAGE" for viewing the pics?

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
```diff
@@ -0,0 +1,385 @@
...
+    private void showPermissionErrorDialog()
+    {
+        AlertDialog.Builder builder;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            builder = new AlertDialog.Builder(this, android.R.style.Theme_Material_Dialog_Alert);
+        } else {
+            builder = new AlertDialog.Builder(this);
+        }
+        builder.setTitle("Permission error")
+                .setMessage("You did not have authorized the app to use internal storage. The app will be closed.")
```
PrandiniUniPD, line 372: In order to avoid hard coding and to favourite the translations, could a good idea be putting these strings to the strings.xml file?

PrandiniUniPD (commented):
Dear Stefano,
I've reviewed your code as requested.
My review would be a sort of talking about your code with the scope to learn together. So I'm pleased you if you'll take every question and every comment with this vision of the things: every comments, mine and yours, are an opportunity to learn together.
I like that you use the fragments because they are more flexible than an Activity without the use of them and I very like the attention about the stack of the fragments.
My comments are about my doubts, I hope they are useful.

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
stefanoromanello: Yes, I left Gallery Activity because is what this activity is about. More details are in the code. I will update it

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
stefanoromanello: Oops, I wrongly wrote the comment, I will fix this

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
stefanoromanello: Yes, is as you said, I will Include a description

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
stefanoromanello: I think that "hidden status" means only if an object is hidden or not. I will explain more clearly

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
stefanoromanello: They are not the same because in I show/hide the activity's buttons. If I want to create a method I need to create it with 3 parameters so I opted for this method because I also get a more clear way to understand the settings of the actionbar are in each fragment

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
stefanoromanello: is used from the GalleryManager on the click of a card from the cardAdapter. I've created the fragment here because the idea is to put all the UI elements for the gallery in GalleryActivity

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
stefanoromanello: Yes WRITE_EXTERNAL_STORAGE is for deleting the images and I don't need to ask for READ_EXTERNAL_STORAGE because android.permission.WRITE_EXTERNAL_STORAGE will automatically add android.permission.READ_EXTERNAL_STORAGE to your APK at build time

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
PrandiniUniPD: Ops, I've seen bad, thanks for the clarification.

app/src/main/java/unipd/se18/ocrcamera/GalleryActivity.java
PrandiniUniPD: What is the advantage to include the implementation of plus than one public fragment class here?
There are other ways to organize the work, such as in packages or in modules if the scope is to isolate some specific classes from each others.

PrandiniUniPD:
The review is a chance to learn together and to do better code.
Thanks to have worked with me.

Checklist:
* Comments about the code
* Explanations about the method or the code in general
* Modularity of the code
* Readability of the code
* Functionality relativeness to Android tools

PR #104, opened 2019-01-10
Codereview #2 of 8 from 3

leopardo-rossi-developer:
With my group we're developing a forum for the app, where users can read useful information/article published by other users and let them know (using comments and likes) if they appreciate what is written. All this information is stored into a DB handled thanks to Firebase Firestore API.

In particular the class you have to review is the one that manages:
1- Posts download from DB.
2- Like addition to a post.


ShowPosts_VM.java
```diff
@@ -0,0 +1,206 @@
...
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.Map;
+
+import unipd.se18.ocrcamera.forum.DatabaseManager;
+import unipd.se18.ocrcamera.forum.R;
+import unipd.se18.ocrcamera.forum.RequestManager;
+import unipd.se18.ocrcamera.forum.models.Post;
+
```
thomasporro, line 30: According to our style document: 

> If there are both static and non-static imports, a single blank line separates the two blocks. There are no other blank lines between import statements.

ShowPosts_VM.java
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * According to the MVVM (Model-View-ViewModel) architecture this class contains all fragment ShowPosts' logic.
+ * This architecture ensures that the code is more testable and organised than the classic approch to put everything
+ * in the activity or fragment class. 
+ * In particular the methods that are implemented here are used to get posts from forum and to handle users' likes. The
```
thomasporro, line 36: Just a distraction error: users' instead of user's

ShowPosts_VM.java
```diff
@@ -0,0 +1,206 @@
...
+                        //At this point, if no exception are thrown, a post object can be built and stored into the
+                        //specific array
+                        posts.add(new Post(postID, postTitle, postMessage, format.parse(postDate), likes, comments, postAuthor));
+                    }
+                    catch (ParseException e)
+                    {
+                        //If an error occurs while converting the post's date an error message is logged to console and 
+                        //the failure UI listener is triggered with a explanation message for the user
+                        Log.d(LOG_TAG, e.getMessage());
+                        if (getPostListener != null){ getPostListener.onGetPostFailure(context.getString(R.string.requestFailedMessage)); }
```
thomasporro, line 142: I think it's a my problem beacause I've never used the listeners but I don't understand where the getPostListener is initialized.

ShowPosts_VM.java
```diff
@@ -0,0 +1,206 @@
...
+        //Definition of the listeners that will be triggered when the db interrogation is finished.
+        //In particular two listeners are defined: one for a successful db interrogation, the other in case of failure
+        DatabaseManager.Listeners listeners = new DatabaseManager.Listeners();
+
+        listeners.successListener = new OnSuccessListener() {
+            @Override
+            public void onSuccess(Object o) {
+                //If the like has correctly been added to the db the success listener is triggerd
+                //(In this case there's an empty string as parameter because there's no specific message to show to the user)
+                if (addLikeListener != null) { addLikeListener.onAddLikeSuccess(""); }
```
thomasporro, line 174: The same thing as above

thomasporro (commented):
The code is really nice to read. The only thing, you might follow more the style doc, for example the column limit of 100 characters. I added a comment in a line that I don't understand too much

PR #106, opened 2019-01-10
codereview#2 of 11 from 9

NicoloCervo:
the code to review is the class AllergenManager (last commit)


app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
```diff
@@ -0,0 +1,254 @@
...
+     */
+    public ArrayList<Allergen> checkForSelectedAllergens(Ingredient ingr) {
+
+        Map<String, ?> selectedAllergens = sp.getAll();
+        Allergen allergen = new Allergen();
+        ArrayList<Allergen> foundSelectedAllergens = new ArrayList<>();
+
+        // For every selected allergen
+        for(Map.Entry<String, ?> entry : selectedAllergens.entrySet()) {
+            // Check if the ingredient has one of it's inci names or it's common name(in the description)
```
AlbVal97, line 162: I couldn't understand the meaning of this comment: may it have been something like: "Checks if the ingredient description contains one of its inci or common names"?

AlbVal97 (commented):
The code is well written: there is no hardcoding, style rules are applied, exceptions are caught, specifications are complete and there is plenty of comments. The only thing I have pointed out in your code is a line of comment which I couldn't understand, but t's likely to be a matter of language. It would be also good if you explained at the beginning the difference between "allergens" and "selectedAllergens" because it's not clear at all.

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
NicoloCervo: yeah, my fault, it's refers to the allergen so it would be "Check if the ingredient has one of the allergen's inci names or it's common name(in the description)". 
I kinda wrote it as a continuation of the previous comment.
description means the ingredient's description if it also wasn't clear.

AlbVal97: Ok, now I understand better what you meant, thank you for clarification!

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
```diff
@@ -0,0 +1,254 @@
...
+
+/**
+ * Class that contains the methods to read the allergens from a file, create an allergen list,
+ * find the allergens in the recognized ingredients from the OCR.
+ */
+class AllergensManager {
+
+    private final int COMMON_NAME = 0;
+    private final int INCI_NAMES = 1;
+
```
AlbVal97, line 26: I couldn't understand the usage of these constants until I have seen them in usage below, so I would suggest to give them more specific names, such as "COMMON_NAME_CSV_INDEX" and "INCI_NAMES_CSV_INDEX". May it be a sensible change in your opinion?

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
```diff
@@ -0,0 +1,254 @@
...
+    /**
+     * constructor, accesses the sharedPreferences file and reads the allergens database from res
+     * @param cntxt context of the app/calling activity
+     * @author Cervo Nicolò
+     */
+    AllergensManager(Context cntxt) {
+
+        context = cntxt;
+        String selectedAllergensSP = cntxt.getString(R.string.selected_allergens_sharedPref);
+        sp = context.getSharedPreferences(selectedAllergensSP, Context.MODE_PRIVATE);
```
AlbVal97, line 41: I don't understand why the Context object is referred both as "context" and "cntxt" in the lines above. Is there a particular reson?

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
```diff
@@ -0,0 +1,254 @@
...
+
+        String line[];
+        String commonName;
+
+        try {
+            while ((line = csvReader.readNext()) != null) {
+
+                Allergen allergen = new Allergen();
+                String inciNames[];
+
```
AlbVal97, line 65: I can't figure out why the array inciNames[] hasn't been declared before the try-catch block, together with the other strings which are used in it.

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
```diff
@@ -0,0 +1,254 @@
...
+     * Reads the allergendb.csv file from resources and stores the content in allergensList
+     * @author Cervo Nicolò
+     */
+    private  void readAllergensdb() {
+
+        // Parse through the allergendb.csv with CSVReader
+        InputStream inputStream = context.getResources().openRawResource(R.raw.allergendb);
+        InputStreamReader allergenDbReader = new InputStreamReader(inputStream);
+        CSVReader csvReader = new CSVReader(allergenDbReader);
+
```
AlbVal97, line 56: May it be a good practice to use the close() method to the csvReader when it will be no longer used?

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
```diff
@@ -0,0 +1,254 @@
...
+            }
+
+        } catch(IOException e){
+            e.printStackTrace();
+        }
+    }
+
+    /**
+     * This method returns an list of Allergen objects from allergendb.csv
+     * @return allergensList, the converted ArrayList of Allergen objects
```
AlbVal97: I think you would say "... returns a list..." ;)

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
```diff
@@ -0,0 +1,254 @@
...
+                allergensList.add(allergen);
+            }
+
+        } catch(IOException e){
+            e.printStackTrace();
+        }
+    }
+
+    /**
+     * This method returns an list of Allergen objects from allergendb.csv
```
AlbVal97, line 83: I think you wanted to say "... returns a list..." ;)

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
NicoloCervo: Maybe it would be better to just add a comment explaining what they are, but you are right.

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
NicoloCervo: I could have used this.context, you are right, will fix

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
NicoloCervo: Originally it was there, but for some reason it after some iterations it would create objects with the inciNames array partially empty (and seemed to keep the largest size of the previous iterations), the only solution i could think of was re-declaring it every time to make sure it wouldn't act weird, it ended up working but I still am not sure if that was the reason.

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
NicoloCervo: You are right.

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
```diff
@@ -0,0 +1,254 @@
...
+        return foundSelectedAllergens;
+    }
+
+    /**
+     * Checks an ArrayList of ingredients for allergens.
+     * @param ingrList, List of ingredients to check.
+     * @return foundAllergens, ArrayList of allergens found in ingrList.
+     * @author Nicolò Cervo (g3)
+     */
+    public ArrayList<Allergen> checkListForAllergens(List<Ingredient> ingrList) {
```
AlbVal97, line 183: I noticed that this method and the previous checkForAllergens() do basically the same operations to different parameters, so I think it would be much better to simply overload the method checkForAllergens() by using two distinct signatures. To a class user perspective it would be less confusing and more correct in a "black box" approach.

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    /**
+     * Checks an ArrayList of ingredients for selected allergens.
+     * @param ingrList, List of ingredients to check.
+     * @return foundAllergens, ArrayList of selected allergens found in ingrList.
+     * @author Nicolò Cervo (g3)
+     */
+    //Todo: check if this method is correct, used int GalleryManager onBindViewHolder
+    public ArrayList<Allergen> checkListForSelectedAllergens(List<Ingredient> ingrList) {
```
AlbVal97, line 199: Similarly to the previous comment, I would suggest you to overload the method checkForSelectedAllergens()

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
```diff
@@ -0,0 +1,254 @@
...
+
+                // set the fields
+                allergen.setCommonName(commonName);
+                allergen.setInciNames(inciNames);
+
+                allergensList.add(allergen);
+            }
+
+        } catch(IOException e){
+            e.printStackTrace();
```
AlbVal97, line 78: While it's good as a diagnostic tool, it surely doesn't help in case of exception. What can be done in common cases such as the allergens csv is only partially filled or empty?

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
AlbVal97: Ok, then I would suggest to futher investigate about this issue.
Another good practice would be to declare and initialize both the variables inciNames[] and commonName at the same time to avoid them being used before initialization, that would lead to "catastrophic" effects in case of small refactorings.

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
```diff
@@ -0,0 +1,254 @@
...
+    /**
+     * Check if an ingredient contains a selected allergen common name or inci name (stored in SharedPreferences sp)
+     * @param ingr, ingredient to check for allergens
+     * @return the allergen found in the ingredient, empty ArrayList if no allergen is found
+     * @author Nicolò Cervo (g3)
+     */
+    public ArrayList<Allergen> checkForSelectedAllergens(Ingredient ingr) {
+
+        Map<String, ?> selectedAllergens = sp.getAll();
+        Allergen allergen = new Allergen();
```
AlbVal97, line 157: I would suggest to declare this variable "allergen" into the "if" block inside the second "for" cycle, in order to allocate its memory only if it's actually needed,

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
```diff
@@ -0,0 +1,254 @@
...
+ */
+class AllergensManager {
+
+    private final int COMMON_NAME = 0;
+    private final int INCI_NAMES = 1;
+
+    private ArrayList<Allergen> allergensList = new ArrayList<>();
+
+    private Context context;
+    private SharedPreferences sp = null;
```
AlbVal97, line 30: I don't understand why this variable is actually initialized as null, considering the fact that this particular value isn't used after that if not to be overwritten inside the constructor. I think in this case it would be better not to initialize it at all.

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
```diff
@@ -0,0 +1,254 @@
...
+
+import unipd.se18.ocrcamera.inci.Ingredient;
+
+
+
+/**
+ * Class that contains the methods to read the allergens from a file, create an allergen list,
+ * find the allergens in the recognized ingredients from the OCR.
+ */
+class AllergensManager {
```
AlbVal97, line 22: Have you considered to define an interface to implement before filling the class with methods? Because as for now it's not abstract at all while it certainly could be.

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
AlbVal97: Yes, yours is definitely an easier and more effective solution

app/src/main/java/unipd/se18/ocrcamera/AllergensManager.java
AlbVal97: Sorry for not having been clear enough: I referred to the fact that getString is called on "cntxt" while getSharedPreferences is on "context" instead

NicoloCervo: You are right, I didn't even realize that, there is no reason, will fix

PR #107, opened 2019-01-10
Codereview #2 of 7 (Prandini) from 1 (Garrido)

PrandiniUniPD:
Dear Oscar,
Here you the code for the review.
It's a class that represents the ViewModel part of the Model - View - ViewModel architecture.
Its work is basically adding a post to an hosted forum and notify processing events to the View.

In order to do this it uses:
- [a class that manages the requests from the server](https://github.com/PrandiniUniPD/OCRCamera/blob/5c222c8388aa03353e4f37d0eda7966faafa425d/forum/src/main/java/unipd/se18/ocrcamera/forum/RequestManager.java) (separated because it's shared for managing different requests in the same class)
- and the [model class](https://github.com/PrandiniUniPD/OCRCamera/blob/5c222c8388aa03353e4f37d0eda7966faafa425d/forum/src/main/java/unipd/se18/ocrcamera/forum/models/Post.java) that represents a post

I hope to improve this code with eventual your suggestions.
Thanks in advance for your time.

P.S.: The review is only for the AddPost_VM.java class but I've written some instrumented test and some junit test for this class, I'll share the links, if you think that they are useful for making more clear my code, you can see them:
- [Instrumented tests](https://github.com/PrandiniUniPD/OCRCamera/blob/discussionForum/forum/src/androidTest/java/unipd/se18/ocrcamera/forum/AddPost_VMInstrumentedTest.java)
- [jUnit tests](https://github.com/PrandiniUniPD/OCRCamera/blob/5c222c8388aa03353e4f37d0eda7966faafa425d/forum/src/test/java/unipd/se18/ocrcamera/forum/AddPost_VMTest.java)

### DIFF
--- a/forum/src/main/java/unipd/se18/ocrcamera/forum/viewmodels/AddPost_VM.java
+++ b/forum/src/main/java/unipd/se18/ocrcamera/forum/viewmodels/AddPost_VM.java
@@ -1,0 +1,320 @@
+package unipd.se18.ocrcamera.forum.viewmodels;
+
+import android.content.Context;
+import android.util.Log;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.ArrayList;
+import java.util.Date;
+
+import unipd.se18.ocrcamera.forum.R;
+import unipd.se18.ocrcamera.forum.RequestManager;
+import unipd.se18.ocrcamera.forum.models.Post;
+
+/**
+ * ViewModel class for adding a post to the forum
+ * (architecture used: Model (the Post) - View (the adding post view) - ViewModel (the worker class))
+ * @see <a href="https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel">
+ *     Model–view–viewmodel</a>
+ * @author Pietro Prandini (g2)
+ */
+public class AddPost_VM implements AddPostsMethods {
+    /**
+     * String used for logs
+     */
+    private static String TAG = "AddPost_VM -> ";
+
+    /**
+     * Key for requesting to add a new post
+     * (used by the server that hosts the forum)
+     */
+    public static final String KEY_ADD_POST_REQUEST = "c";
+
+    /**
+     * Key for indicating the content of the post
+     * (used by the server that hosts the forum)
+     */
+    public static final String KEY_JSON_POST_CONTENT = "jPost";
+
+    /*
+    The next listener is provided by this class and it will be implemented by the View class
+    for receiving notifications about the work of this ViewModel class.
+     */
+
+    /**
+     * Listener useful for communicating with the View
+     * @see <a href="https://docs.oracle.com/javase/tutorial/uiswing/events/index.html">
+     *     Writing Event Listeners</a>
+     * @author Pietro Prandini (g2)
+     */
+    public interface addPostListener {
+        /**
+         * Notifies when a post is correctly added
+         * @param response The response of the server
+         */
+        void onPostAdded(String response);
+
+        /**
+         * Notifies a connection problem to the network
+         * @param error The error parsed
+         */
+        void onConnectionFailed(String error);
+
+        /**
+         * Notifies a failure of a sending parameters process addressed to the server
+         * @param error The error parsed
+         */
+        void onParametersSendingFailed(String error);
+
+        /**
+         * Notifies the VM has received some not valid parameters
+         * (title, message and author null or empty String)
+         * @param error The String that describes the error
+         */
+        void onNotValidParameters(String error);
+
+        /**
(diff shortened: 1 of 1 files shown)


Nintenking:
I spent a few hours to read the class you gave me to review and the enclosed documentation; as expected this is a whole new level of writing code, so I'm thankful I got to read it to improve my own skills.
I'll just ask a couple of things, they're more of a lack of understanding of mine due to my inexperience than actual errors of code.

forum/src/main/java/unipd/se18/ocrcamera/forum/viewmodels/AddPost_VM.java
```diff
@@ -0,0 +1,320 @@
...
+     */
+    private Boolean areParametersValid(String title, String message, String author) {
+        // Checks if the Strings are not null and not empty ones
+        return (title != null && !title.equals(""))
+                && (message != null && !message.equals(""))
+                && (author != null && !author.equals(""));
+    }
+
+    /*
+    The next two methods are public and static to favourite the testing
```
Nintenking, line 239: By "testing" you mean UnitTesting and Mocking, right? 
I don't quite get why the following methods are public, by what they do I would think they would better be private. But I'm really not used to write testing of the code, so I'm hearing you out!

forum/src/main/java/unipd/se18/ocrcamera/forum/viewmodels/AddPost_VM.java
```diff
@@ -0,0 +1,320 @@
...
+     * {@link RequestManager#sendRequest(Context, ArrayList)}.
+     * @param context The reference of the activity/fragment that calls this method
+     * @param title The new post's title
+     * @param message The new post's message
+     * @param author The new post's author
+     * @author Pietro Prandini (g2)
+     */
+    @Override
+    public void addPostToForum(final Context context, String title, String message, String author) {
+        Log.i(TAG,"addPostToForum");
```
Nintenking, line 180: This is one of the various logs you use, so I'll take it as an example; obviously, I use them as well to read in the logcat what is happening in a determined place of my code; my question is: it is correct to compile them also in the final version of the app? Or should they be removed at some point of the developing process?

PrandiniUniPD:
> I spent a few hours to read the class you gave me to review and the enclosed documentation; as expected this is a whole new level of writing code, so I'm thankful I got to read it to improve my own skills.
> I'll just ask a couple of things, they're more of a lack of understanding of mine due to my inexperience than actual errors of code.

Don't worry, we are here for learning together, both, you and me, so feel free to ask me the things that are not clear.

forum/src/main/java/unipd/se18/ocrcamera/forum/viewmodels/AddPost_VM.java
PrandiniUniPD: In my experience the logs don't compromise the user experience and at the same time they support the developers, so I usually left them but there are some levels of logs described also in the [android API](https://developer.android.com/reference/android/util/Log) and the "verbose" level it could be considered only for using under development.

forum/src/main/java/unipd/se18/ocrcamera/forum/viewmodels/AddPost_VM.java
PrandiniUniPD: Yes, for testing i've meant junit tests and instrumented tests of android (in the instrumented test I've used some mocked object but it's not relevant) as I've written in the description of this pull request. Should I write more details in this comment?

The story about that is this: I'd to respect an interface that has imposed to me the parameters and the return type of the method addPostToForum but at the same time I'd to split that method in plus that one part for granting a certain level of readability.
So I'd use the next two method for preparing the request (an arrylist of parameters) and for preparing the String of the post in a JSON format.
If these two methods would be private I wouldn't had accessed to them for testing them but I've thought that was important test to avoid the risk to send something wrong to the server and in this sense I've preferred to test these method. But these two methods only prepare a request (ones return a JSON string and ones return an arraylist of parameters), the real senders is addPostToForum.

Nintenking: Ok, the code is already commented enough, no need to add more details.
Regardless I have now understood more clearly what you meant to do. Thank you

Nintenking:
I would say we are done here, congrats on the great work you put in this project.
Thank you for your time, I'm now closing the pull request👌

PrandiniUniPD:
> I would say we are done here, congrats on the great work you put in this project.
> Thank you for your time, I'm now closing the pull requestok_hand

Thank you too, all together we can learn and we can do a good job.
Have a nice day.

PR #110, opened 2019-01-10
Codereview #2 of 6 from 14

GiovanniFurlan:
Hi Stefano Romanello,
i ask you to review my code please.
The code is in the file ManualTestOnSinglePhoto.
This class receive an angle by user. Then rotate the last image taken by this angle and use OCR on it. Then compare the text from normal and rotated photo and gave back similarity between text.
If you have some suggestion please tell me.
Thanks you.


app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /**
+     *Analise the given string and get the right warnings and add them on waring variable
+     * warning are found
+     * @param text String to analise
+     * @param length int of length of starting string
+     * @return  warnings
+     */
+    private String findWarning(java.lang.String text, int length) {
```
stefanoromanello: Good to implement a warning but if I get more text from the rotated image I don't get any warning/message?
As is I get warning only if the new rotated image extract less characters than the original.

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
```diff
@@ -0,0 +1,212 @@
...
+            }
+        });
+    }
+
+    /**Based on the text received and startingOCRTText, analise information like confidence
+     * and show them on textView
+     * @param text string got from OCR
+     * @modify degreeTextView, confidenceTextView, warningTextView and foundTextView
+     */
+    private void SetResult(String text){
```
stefanoromanello: Maybe you were wrong to write but the methods should start with a lowercase letter.
Also I suggest to put the content of this method inside onTextRecognized, in this way all the UI contents are in the same place and you don't need to create 5 private objects for the UI in this class. What do you think?

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
```diff
@@ -0,0 +1,212 @@
...
+            SetResult(text);
+        }
+
+        @Override
+        public void onTextRecognizedError(int code) {
+            /*
+             Text not correctly recognized
+             Start CameraActivity to take a new photo
+             */
+            Intent i = new Intent(ManualTestOnSinglePhoto.this, CameraActivity.class);
```
stefanoromanello: I suggest to use a more correct name for this variable like cameraActivity

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
```diff
@@ -0,0 +1,212 @@
...
+
+
+        //Get image path and text of the last image from preferences
+        SharedPreferences prefs = this.getSharedPreferences("prefs", MODE_PRIVATE);
+        String pathImage = prefs.getString("imagePath", null);
+        startingOCRText = prefs.getString("text", null);
+
+        final Bitmap photo = BitmapFactory.decodeFile(pathImage);
+
+        Button fab = findViewById(R.id.btnConfirm);
```
stefanoromanello: I suggest a better name for this button object like buttonAction or a better name for describe the object

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
```diff
@@ -0,0 +1,212 @@
...
+     */
+    private String findWarning(java.lang.String text, int length) {
+        String warnings = "";
+        final int firstBound=10;
+        final int secondBound=20;
+        final int thirdBound=30;
+        final int toleranceValue=4;
+
+
+        if (text.equals("")) {
```
stefanoromanello: Good to implement a warning but maybe I understood wrongly this part. Why are you only sending the warning if the text extracted is less than 30? Also why are you using these 3 bounds? Shouldn't be better to do like (length-text.lenght) and get the length difference? In this way I can understand if I get more or fewer letters. If the result is positive I've lost letters, If its negative I've obtained more letters.
Because I don't see the utility to know if the new extracted text is less than 30 or 20 characters, with this class I want to know if the rotation improves or aggravate the image recognition.
Tell me what you think

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
stefanoromanello: I suggest to change the name of these 2 parameters. A possible solution could be newExtractedText and originalTextLength

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * Modification variable
+     */
+    private int angleRotation;
+
+    /**
+     * UI
+     */
+    private EditText degreeEditText;
```
stefanoromanello: degreeEditText is used only in onCreate. Is there a reason to keep it outside the onCreate?

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
```diff
@@ -0,0 +1,212 @@
...
+    @Override
+    public void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        //UI initialization
+        setContentView(R.layout.activity_manual_test);
+        degreeEditText = findViewById(R.id.editText1);
+
+
+        //Get image path and text of the last image from preferences
+        SharedPreferences prefs = this.getSharedPreferences("prefs", MODE_PRIVATE);
```
stefanoromanello: I suggest to add the name of the SharedPreferences to the xml resource for the Strings

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
GiovanniFurlan: Just to have all the variable at the beginning  of the class. Do u think it's better that way or create a new variable in the method where it is used, if it's used only in this method?

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
GiovanniFurlan: You are right

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
GiovanniFurlan: Yes thanks for the advice

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
GiovanniFurlan: Yeah, it's the only button but anyway this would be better

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
GiovanniFurlan: Yes just distraction error

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
GiovanniFurlan: Thanks for this good advice

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
GiovanniFurlan: Your solution is good, i would think about this new implementation

app/src/main/java/unipd/se18/ocrcamera/ManualTestOnSinglePhoto.java
stefanoromanello: Better to create a variable inside the method if its used only there

PR #112, opened 2019-01-10
Codereview #2 of 0 (Fasan) from 13 (Pham)

Giofasan:
Hi Francesco,
I worked on 4 classes: Inci, PreProcessing, PreProcessingMethods and Tesseract.
I kindly ask you to review my code and in case you give me some ideas about what to improve.
Thank you.
(Edit) By mistake I also entered the tesseract class which is not to be revised. I apologize.


app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * @author Giovanni Fasan(g1)
+     * @param inputStream of the R.raw.database
+     * @return void
+     * Load every Inci ingredients to arrayList
+     */
+    private void loadDB(InputStream inputStream){
+        //define 0 the name of the ingredients
+        int name=0;
```
frankplus, line 54: if this is a constant it would be better if you declare it final. This way you can avoid to accidentally change it somewhere else in the code and it would also be more clear when other programmers read the code.

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+     */
+    public void findIngredientsList(String text) {
+        listIngredientFound = new ArrayList<String>();
+        //remove every \n and every :
+        //split by Ingredients to reduce the number of words to be analyzed
+        String[] noEnd = text.replaceAll("\n", " ").replaceAll(":", " ").split("Ingredients");
+        if (noEnd.length==2){
+            //if I have a length of the vector is 2, I take the second part of the vector, that is the one where there is the list of ingredients
+            String [] word = noEnd[1].split(",");
+            Log.d("inci", "=2");
```
frankplus, line 124: this is not an explicative log (if it is temporary please remove it)

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * @author Giovanni Fasan(g1)
+     * @param text String in which you have to find the ingredients
+     * loads all the ingredients inci found in the ArrayList listIngredientFound
+     */
+    public void findIngredientsList(String text) {
+        listIngredientFound = new ArrayList<String>();
+        //remove every \n and every :
+        //split by Ingredients to reduce the number of words to be analyzed
+        String[] noEnd = text.replaceAll("\n", " ").replaceAll(":", " ").split("Ingredients");
```
frankplus, line 120: please use more explicative variable names

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * @author Giovanni Fasan(g1)
+     * @return String with ingredients list of the label
+     * Concatenate all the ingredients found in a string. It puts the ingredidients in bold
+     */
+
+    public String ingredientsFoundToString() {
+        String inci="";
+        for (int j=0; j<this.listIngredientFound.size(); j++){
+            inci = inci + "<b>" + listIngredientFound.get(j) + "</b>; ";
```
frankplus, line 83: for multiple string concatenation in a loop such as this it would be better to use the StringBuilder.

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+
+        return inci;
+    }
+
+    /**
+     * @author Giovanni Fasan(g1)
+     * @param text String in which you have to find the ingredients
+     * loads all the ingredients inci found in the ArrayList listIngredientFound
+     */
+    private void findIngredientsList_old(String text) {
```
frankplus, line 94: underscore in the method name does not comply the style document. It should be camelCase (if this is temporary please remove it)

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,369 @@
...
+     * Detect if the image is bright
+     * @param imageMat the image we want to detect the brightness
+     * @return 0 if image is neither too bright nor too dark,
+     *         1 if image is too bright,
+     *         2 if image is too dark.
+     * @author Thomas Porro(g1), Giovanni Fasan(g1), Leonardo Pratesi(g1)
+     */
+    private int isBright(Mat imageMat){
+        //Converts the image into a matrix
+        Mat brightnessMat = new Mat();
```
frankplus, line 172: how does "new Mat()" convert an image into a matrix? isn't imageMat already a matrix?

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,369 @@
...
+        }
+        //Calculate the percentage of the brightness
+        double brightness = sum.val[0]/((Math.pow(2,bit)-1)*brightnessMat.rows()*brightnessMat.cols())*2;
+
+        Log.d(TAG, "Brightness:"+brightness);
+
+        /*Bounds to define if the image is dark or bright.
+          The values were decided on the basis of various tests*/
+        double upperBound = 0.9;
+        double lowerBound = 0.4;
```
frankplus, line 219: these variables could be final

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+
+    /**
+     * Change the brightness of the image into an optimal one
+     * @param image the image we want to modify the brightness
+     * @return the image with the modified brightness
+     * @author Thomas Porro(g1), Giovanni Fasan(g1), Oscar Garrido(g1)
+     */
+    private Bitmap editBright(Bitmap image){
```
frankplus, line 237: I suggest to call this method "adjustBrightness" because if it simply changes the brightness then we need some parameters to specify how it should be changed. Instead this method automatically improves the brightness.

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,369 @@
...
+        switch (isBright(bright)) {
+            case 1: //If the image is too bright
+                Log.d(TAG, "Case==1 ==> Too bright");
+                //Darkens the colour's brightness until it's in an optimal value
+                for(double changeBrightness=0; changeBrightness!=-240; changeBrightness-=15){
+                    //Converts an array to another data type with optional scaling.
+                    bright.convertTo(modifiedMat, -1, 1, changeBrightness);
+                    if(isBright(modifiedMat)==0){
+                        try{
+                            return IPUtils.conversionMatToBitmap(modifiedMat);
```
frankplus, line 259: I think that you shoudn't put a return statement in a for. it's generally better to use one ending  return statement because it is easier to figure out where the method will exit.

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,369 @@
...
+                            Log.e(TAG, error.getErrorMessage());
+                            return image;
+                        }
+                    }
+                }
+                break;
+
+            case 0: //Image is neither too bright nor too dark
+                Log.d(TAG, "Case==0 ==> Perfect image");
+                return image;
```
frankplus, line 288: this return statement isn't necessary

app/src/main/java/unipd/se18/ocrcamera/Tesseract.java
```diff
@@ -0,0 +1,103 @@
...
+
+
+class Tesseract extends Application implements TextRecognizer{
+
+    //Tesseract API
+    private TessBaseAPI mTess;
+    //The datapath to the folder where the file trained data of tesseract is located
+    String datapath;
+    //language chosen to recognize the text
+    String language;
```
frankplus, line 26: variables in classes should always be private in OOP! (edit: sorry I didn't read that I shoudn't review Tesseract)

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+            }
+        }
+
+    }
+
+
+    /**
+     * @author Giovanni Fasan(g1)
+     * @param text String in which you have to find the ingredients
+     * loads all the ingredients inci found in the ArrayList listIngredientFound
```
frankplus, line 114: In order to comply javadocs you should put the description first, followed by block tags. (see "How to Write Doc Comments for the Javadoc Tool"  https://www.oracle.com/technetwork/java/javase/index-137868.html)

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,369 @@
...
+        return rotateBitmap(image, angle);
+    }
+
+
+    /**
+     * Detect if the image is bright
+     * @param imageMat the image we want to detect the brightness
+     * @return 0 if image is neither too bright nor too dark,
+     *         1 if image is too bright,
+     *         2 if image is too dark.
```
frankplus, line 167: instead of using 0,1,2 you could declare an enumaration such as `private enum Brightness {GOOD, TOO_BRIGHT, TOO_DARK`} (see https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html)

frankplus:
Hi Giovanni, these are some things that needs to be checked in my opinion. However I would be glad to hear your opinion as well.

app/src/main/java/unipd/se18/ocrcamera/Inci.java
Giofasan: Yes, you are perfectly right. Honestly, I am surprised that I did not put it.

app/src/main/java/unipd/se18/ocrcamera/Inci.java
Giofasan: Thanks a lot. I did not know this class. I will definitely use it

app/src/main/java/unipd/se18/ocrcamera/Inci.java
Giofasan: Yes I had thought about it but I would have had three more variables to use them once.

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
Giofasan: We added it, but unfortunately I had loaded the code before this addition

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
Giofasan: you are right

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
Giofasan: Thank you for your suggestion. We will modify the name

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
Giofasan: you're probably right. I will fix the code to make the return out of the for

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
Giofasan: it's a wrong comment, sorry. new Mat () creates a new instance of a Mat object.

PR #113, opened 2019-01-10
Codereview #2 of 0 (Fasan) from 7 (Prandini)

Giofasan:
Hi Pietro,
I worked on 4 classes: Inci, PreProcessing, PreProcessingMethods and Tesseract.
I kindly ask you to review my code and in case you give me some ideas about what to improve.
Thank you.
(Edit) By mistake I also entered the tesseract class which is not to be revised. I apologize.


app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+
+import android.util.Log;
+
+import java.io.BufferedReader;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+
+public class Inci {
```
PrandiniUniPD, line 11: It's nice and it's useful introducing what this class does to the developer(s), for example with some javadoc comments immediately before this line.
With the javadoc is also possible put some useful links useful for understanding the context of the work, for example with a line like this "@see \<a href\=\"[https://en.wikipedia.org/wiki/International_Nomenclature_of_Cosmetic_Ingredients](https://en.wikipedia.org/wiki/International_Nomenclature_of_Cosmetic_Ingredients)">International Nomenclature of Cosmetic ingredients (INCI)\<\/a\>
What do you think about this?

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+
+    //list where save the ingredients
+    private ArrayList<String> listInci;
+    //list where save the description of ingredients
+    private ArrayList<String> listInciDescription;
+    //list where save the ingredients found
+    private ArrayList<String> listIngredientFound;
+
+    /**
+     * @author Giovanni Fasan(g1)
```
PrandiniUniPD, line 21: In order to emphasize the method, I usually prefer firstly putting the description of the method, @param, @return and @throws and then the @author on the javadoc comments. What do you think about this?

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+     * with loadDB it load Inci ingredients to arraylist
+     */
+    public Inci(InputStream inputStream) {
+        listInci = new ArrayList<String>();
+        listInciDescription = new ArrayList<String>();
+        listIngredientFound = new ArrayList<String>();
+        loadDB(inputStream);
+    }
+
+    public ArrayList<String> getListInci(){
```
PrandiniUniPD, line 33: It's a good practice making javadoc comments to all the methods. What do you think about this?

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+        //define 0 the name of the ingredients
+        int name=0;
+        //define 1 the description of the ingredients
+        int description=1;
+        //open Buffer Reader for read the database
+        BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, Charset.forName("UTF-8")));
+        String ingredient;
+        try {
+            //slide every line of the database and add in the appropriate list
+            while ((ingredient = reader.readLine()) != null) {
```
PrandiniUniPD, line 62: The String ingredient of the line 59 seems to be not useful outer this while loop, is it right?
``for(String ingredient = reader.readLine(); ingredient != null; ingredient = reader.readLine()) { ``
   ``   ... ``
``}``
Could this other implementation more elegant?

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+                String[] inci=ingredient.split(";");
+                listInci.add(inci[name].trim());
```
PrandiniUniPD, line 64: I'm sorry but I don't know how the inci database is structured, maybe a link about this or a short description could be useful to motivate your code?
Anyway I think that the name and the description are biunivocally linked, is it right?
If yes, could a good idea be to use an ArrayList of String[] and add to it array of strings like ("name", "description")?
In this way it's possible to add or to remove ingredient (and the relative description) without think about updating indexes between more than one lists.

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+                //TODO listInciDescription
+                //listInciDescription.add(inci[description].trim());
+            }
+            Log.d("inci", "create listInci");
+        }catch (Exception e){
+            e.printStackTrace();
```
PrandiniUniPD, line 70: "If this try-catch fails what's happen? Is the case to advise the user about the problem or is the case the engine restart automatically with a recover mode or wich is the case? How could the code react to a possible fail?"
Could these question be useful to make a good code?
Usually it's good specifing the exeptions, maybe you are referring to "UnsupportedEncodingException" for the execution of the InputStramReader and to "IOExeption" for the execution for the BufferedReader and not to all the possible exeptions.
Could a good idea be to specify the exeptions and the reaction about every single specific exeption in order to make a more solid software?

PrandiniUniPD, line 70: "If this try-catch fails what's happen? Is the case to advise the user about the problem or is the case the engine restart automatically with a recover mode or wich is the case? How could the code react to a possible fail?"
Could these question be useful to make a good code?
Usually it's good specifing the exeptions, maybe you are referring to "UnsupportedEncodingException" for the execution of the InputStramReader and to "IOExeption" for the execution for the BufferedReader and not to all the possible exeptions.
Could a good idea be to specify the exeptions and the reaction about every single specific exeption in order to make a more solid software?

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+
+        return inci;
+    }
+
+    /**
+     * @author Giovanni Fasan(g1)
+     * @param text String in which you have to find the ingredients
+     * loads all the ingredients inci found in the ArrayList listIngredientFound
+     */
+    private void findIngredientsList_old(String text) {
```
PrandiniUniPD, line 94: For "_old" what do you mean? Is the obsolete implementation? If yes, what is the reason to maintain this method in the code?

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+
+    }
+
+
+    /**
+     * @author Giovanni Fasan(g1)
+     * @param text String in which you have to find the ingredients
+     * loads all the ingredients inci found in the ArrayList listIngredientFound
+     */
+    public void findIngredientsList(String text) {
```
PrandiniUniPD, line 116: This method is void and put the ingredients found from a String to the arraylist named listIngredientFound, is it right?
But I don't see synchronization structures or design patterns used to notify when the work of this method is completed like a listener or design pattern used to observe the behaviour of a livedata like an observer, so I've not so clear how the system works.
What am I missing for understanding?

app/src/main/java/unipd/se18/ocrcamera/Inci.java
```diff
@@ -0,0 +1,148 @@
...
+        listIngredientFound = new ArrayList<String>();
+        //remove every \n and every :
+        //split by Ingredients to reduce the number of words to be analyzed
+        String[] noEnd = text.replaceAll("\n", " ").replaceAll(":", " ").split("Ingredients");
+        if (noEnd.length==2){
+            //if I have a length of the vector is 2, I take the second part of the vector, that is the one where there is the list of ingredients
+            String [] word = noEnd[1].split(",");
+            Log.d("inci", "=2");
```
PrandiniUniPD, line 124: Could a good idea be using a final TAG String instead the tag of the class hard coded here?

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,369 @@
...
+    //Tag used to identify the log
+    private final String TAG = "PreProcessing";
+
+
+    /**
+     * Constructor of the class which initialize the openCV library
+     * @author Thomas Porro (g1)
+     */
+    public PreProcessing() {
+        //TODO verify if the library is correctly loaded
```
PrandiniUniPD, line 40: Why is it only a "to do" and it isn't code this check? Is it not important to do?

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,369 @@
...
+    /**
+     * Rotate the given bitmap with the given angle
+     * @param original The image that we want to corrected
+     * @param degrees The degrees of the angle we want to rotate
+     * @return the rotatedImage. If the image is null return the original image
+     * @author Thomas Porro(g1), Giovanni Fasan (g1), Oscar Garrido (g1)
+     */
+    private static Bitmap rotateBitmap(Bitmap original, double degrees) {
+
+        //TODO not correct because it returns null, maybe we should add a @NonNull
```
PrandiniUniPD, line 130: What is this todo? Have I to review this method?

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,369 @@
...
+        return rotateBitmap(image, angle);
+    }
+
+
+    /**
+     * Detect if the image is bright
+     * @param imageMat the image we want to detect the brightness
+     * @return 0 if image is neither too bright nor too dark,
+     *         1 if image is too bright,
+     *         2 if image is too dark.
```
PrandiniUniPD, line 167: Could a good idea be declaring 3 public static final int for containing and using these 3 values in order to support the use of this method and to support the developing of this method?

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,369 @@
...
+
+        //Changes the format of the matrix
+        Imgproc.cvtColor(imageMat, brightnessMat, Imgproc.COLOR_RGBA2RGB);
+
+        //Obtain 3 different matrix with the 3 elemental colors
+        List<Mat> color = new ArrayList<>();
+        Core.split(brightnessMat, color);
+
+        /*Each color is multiplied with his luminance.
+          For more informarions see https://en.wikipedia.org/wiki/Relative_luminance*/
```
PrandiniUniPD, line 182: If you like there is a method to include link as this in the javadoc comments, in this case it is similar to this:
@see \<a href\="[https://en.wikipedia.org/wiki/Relative_luminance](https://en.wikipedia.org/wiki/Relative_luminance)"\>Relative luminance\</a\>

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,369 @@
...
+                Log.d(TAG, "Case==0 ==> Perfect image");
+                return image;
+        }
+        return image;
+    }
+
+
+    /**
+     * @author Thomas Porro(g1), Oscar Garrido (g1), Giovanni Fasan(g1).
+     * See PreProcessingMethods.java
```
PrandiniUniPD, line 296: I appreciate you for this indication, in general I think that indication for coder or reviewer clarify the code.
If you would like to use the javadoc method to do this, in this case, it's similar to this {@link PreProcessingMethods}

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+
+    /**
+     * Change the brightness of the image into an optimal one
+     * @param image the image we want to modify the brightness
+     * @return the image with the modified brightness
+     * @author Thomas Porro(g1), Giovanni Fasan(g1), Oscar Garrido(g1)
+     */
+    private Bitmap editBright(Bitmap image){
```
PrandiniUniPD, line 237: Could the name of this method be "getBrightOptimizedBitmap" or something similar? With editBright I've a feel like to "manually" edit by parameters the brightness of a Bitmap.

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,369 @@
...
+        //Create a Bitmap with the given matrix, and obtain all the pixels from it
+        Bitmap laplacianImage;
+        try{
+            laplacianImage = IPUtils.conversionMatToBitmap(laplacianMat8Bit);
+        } catch (ConversionFailedException error){
+            Log.e(TAG, error.getErrorMessage());
+            return false;
+        }
+
+        //TODO we have to comment this lines and add getPixels to the builder
```
PrandiniUniPD, line 336: Is this code for testing? Should I review this? Why are there some "todos" that seem important not done sometimes?

PrandiniUniPD (commented):
Dear Giovanni,
I've reviewed your code as requested: Inci.java, PreProcessing.java (only your methods), PreProcessingMethods.java.
My review would be a sort of talking about your code with the scope to learn together. So I'm pleased you if you'll take every question and every comment with this vision of the things: every comments, mine and yours, are an opportunity to learn together.
Thanks in advance.

app/src/main/java/unipd/se18/ocrcamera/Inci.java
Giofasan: it's an excellent idea

app/src/main/java/unipd/se18/ocrcamera/Inci.java
Giofasan: Yes, it's true. I created the getters automatically, which is why I forgot to comment on these methods.

app/src/main/java/unipd/se18/ocrcamera/Inci.java
Giofasan: I preferred to use this mode for the moment as I still have to work on the description. I thank you for the advice, I will put it into practice.

app/src/main/java/unipd/se18/ocrcamera/Inci.java
Giofasan: Yes you are right. I wrote this part of the code badly. I will immediately place the code.

app/src/main/java/unipd/se18/ocrcamera/Inci.java
Giofasan: it's an obsolete code. I left it to check the result with the new function. the function will be canceled

app/src/main/java/unipd/se18/ocrcamera/Inci.java
Giofasan: I forgot to remove it

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
Giofasan: The idea is to later add the impossibility of passing a null value as a parameter

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
Giofasan: We have created an enum class to solve this problem. But unfortunately I had loaded the code before this addition

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
Giofasan: Thank you for your suggestion

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
Giofasan: Yes you are right. We have to change the name of the method

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
PrandiniUniPD: Good idea

PrandiniUniPD:
The review is a chance to learn together and to do better code.
Thanks to have worked with me.

Checklist:
* Comments about the code
* Explanations about the method or the code in general
* Modularity of the code
* Readability of the code
* Functionality relativeness to Android tools

PR #114, opened 2019-01-10
codereview #2 of 12 (Moroldo) from 0 (Fasan)

lucamoroz:
Dear Mr. Fasan,
I hope this message finds you well.
The file involved in this review is SearchResultActivity, an activity that allow the user to manually search for any ingredient and display the ingredient data.
Please feel free to ask for anything about the code.


app/src/main/java/unipd/se18/ocrcamera/SearchResultActivity.java
```diff
@@ -0,0 +1,407 @@
...
+
+    /**
+     * Used to find ingredients from the INCI db with a query
+     */
+    private IngredientsExtractor ingredientsExtractor;
+
+    /**
+     * Size of the dropdown suggestions list
+     */
+    private static final int MAX_DROPDOWN_SUGGESTIONS = 40;
```
Giofasan: In my opinion 40 as Size of the dropdown list is too high. I recommend reducing it to 10/20.

app/src/main/java/unipd/se18/ocrcamera/SearchResultActivity.java
```diff
@@ -0,0 +1,407 @@
...
+
+        //load the complete ingredients list and setup the AutoCompleteTextView
+        new AutoCompleteTVSetup().run();
+
+        //handle the calling intent
+        handleIntent(getIntent());
+    }
+
+    @Override
+    protected void onNewIntent(Intent intent) {
```
Giofasan, line 162: I recommend adding a brief comment on what this function does.

app/src/main/java/unipd/se18/ocrcamera/SearchResultActivity.java
```diff
@@ -0,0 +1,407 @@
...
+     */
+    private void handleIntent(Intent intent) {
+
+        if (Intent.ACTION_SEARCH.equals(intent.getAction())) {
+            //get query text
+            String query = intent.getStringExtra(SearchManager.QUERY);
+
+            //update textview text and set cursor to the end
+            mAutoCompleteTextView.setText(query);
+            mAutoCompleteTextView.setSelection(query.length());
```
Giofasan, line 179: I do not understand what the setSelection function does

app/src/main/java/unipd/se18/ocrcamera/SearchResultActivity.java
```diff
@@ -0,0 +1,407 @@
...
+            mAutoCompleteTextView.setOnItemClickListener(new AdapterView.OnItemClickListener() {
+                @Override
+                public void onItemClick(AdapterView<?> parent, View view, int position, long id) {
+
+                    String ingredientClicked = parent.getItemAtPosition(position).toString();
+
+                    //update TextView and set cursor at the end of the text
+                    mAutoCompleteTextView.setText(ingredientClicked);
+                    mAutoCompleteTextView.setSelection(mAutoCompleteTextView.getText().length());
+                    mSearchButton.performClick();
```
Giofasan, line 258: I suggest you add a comment to explain what the function .performClick () does.
Like this "Call this view's OnClickListener, if it is defined. Performs all normal actions associated with clicking".

app/src/main/java/unipd/se18/ocrcamera/SearchResultActivity.java
```diff
@@ -0,0 +1,407 @@
...
+        private List<String> lastList;
+
+        private IngredientsFilter() {
+            super();
+            lastSize = 0;
+            lastList = getFullIngredientsList();
+        }
+
+        @Override
+        protected FilterResults performFiltering(CharSequence constraint) {
```
Giofasan, line 306: why do you use a CharSequence variable instead of a String variable since then you have to convert the CharSequence variable to String?

app/src/main/java/unipd/se18/ocrcamera/SearchResultActivity.java
```diff
@@ -0,0 +1,407 @@
...
+
+
+    /**
+     * Load all INCI ingredients names, set them as suggestions with a custom filter for the
+     * AutoCompleteTextView and set a suggestion click listener.
+     * The custom filter will show the suggestion in lexicographical order, privileging words that
+     * starts with the constraint query or contains a substring that starts with the constraint
+     * query.
+     */
+    private class AutoCompleteTVSetup implements Runnable {
```
Giofasan, line 215: Great idea to use a thread

app/src/main/java/unipd/se18/ocrcamera/SearchResultActivity.java
lucamoroz: It sets the cursor at the end of the word, so the user can eventually append some text after choosing any suggestion

app/src/main/java/unipd/se18/ocrcamera/SearchResultActivity.java
lucamoroz: I'll add a comment soon, thank you for the report

app/src/main/java/unipd/se18/ocrcamera/SearchResultActivity.java
lucamoroz: Ok thank you for reporting that

app/src/main/java/unipd/se18/ocrcamera/SearchResultActivity.java
lucamoroz: Because I'm overriding this class functions:
https://developer.android.com/reference/android/widget/Filter

PR #116, opened 2019-01-10
Codereview #2 of 13 (Pham) from 6 (Furlan)

frankplus:
Hello @GiovanniFurlan ,
here is my code that needs to be reviewed.
I briefly summarize what I have implemented these days:
- **Inci**: Load INCI database from csv file into a data structure (using openCSV).

- **TextAutoCorrection**: automatic correction of the ocr text using a dictionary of different words taken from INCI DB.

- **IngredientsExtractor**: extraction of a list of ingredients from the ocr text. 
I did two implementations: _NameMatchIngredientsExtractor_ and _TextSplitIngredientsExtractor_. Only the first one is used in the app because during testing I found that it performed much better than the second one. I still use the second one for testing purposes.

- **PhotoTester.buildIngredientsExtractionReport** (do not review other parts of PhotoTester): here the extraction is tested on each picture in the test set and some statistics are calculated.

- **ResultActivity.AsyncIngredientsExtraction** (do not review other parts of ResultActivity): 
    - execute extraction in an asynctask
    - print analyzed ocr text highlighting the ingredients found.
    - display the extracted ingredients  into a list view using _AdapterIngredient_ (also highlighting allergens)

- **IngredientDetailsFragment**: this is a DialogFragment that pops up when an ingredient in the result list is clicked. In this fragment details about the ingredient is shown including an extract from wikipedia and a button for a search on google.
- **IngredientsExtractionTest**: here you can find some Junit tests of the ingredients extraction

p.s. in Utils.java just look at where my name is written

Thanks a lot for your time!
Francesco


app/src/main/java/unipd/se18/ocrcamera/inci/Inci.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Class for Inci database management
+ * The database is in CSV format, using openCSV to parse the file.
+ * For more informations about openCSV:
+ * http://opencsv.sourceforge.net/
+ * @author Francesco Pham
+ */
+public class Inci {
+
+    private static final String TAG = "Inci";
```
GiovanniFurlan: For a better look i advice to add all string in string.xml file in resources and then retrive it when u need. Also better if u want to translate the app in the future

app/src/main/java/unipd/se18/ocrcamera/inci/Inci.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private static final int COL_COSING_REF_NO = 0;
+    private static final int COL_INCI_NAME = 1;
+    private static final int COL_DESCRIPTION = 6;
+    private static final int COL_FUNCTION = 8;
```
GiovanniFurlan, line 28: Could be better to use an ENUM for descrive this 4 final variables

app/src/main/java/unipd/se18/ocrcamera/inci/Inci.java
```diff
@@ -0,0 +1,81 @@
...
+
+        //initialize openCSV reader
+        CSVReader csvReader = new CSVReader(reader);
+
+        //skip first line containing field names
+        try {
+            csvReader.skip(1);
+        } catch (IOException e) {
+            Log.e(TAG, "Error skipping first line");
+        }
```
GiovanniFurlan, line 48: is this try and catch necessary, could u not just skip the first line in the try and catch after that.
If it's necessary could be better to not catch the exception with just a log but find a way to overstep or at least back to the main activity. 
The same advice for the two try-catch after this one

app/src/main/java/unipd/se18/ocrcamera/inci/Inci.java
```diff
@@ -0,0 +1,81 @@
...
+            reader.close();
+            csvReader.close();
+        } catch(IOException e){
+            Log.e(TAG, "Error closing csv reader");
+        }
+
+        return listIngredients;
+    }
+
+}
```
GiovanniFurlan, line 81: Hardcoding at line 23,47,63,67,75

app/src/main/java/unipd/se18/ocrcamera/inci/TextAutoCorrection.java
```diff
@@ -0,0 +1,164 @@
...
+    /**
+     * Each word of the text is searched for a best match in the dictionary and
+     * if there is a good match the word is corrected
+     * @param text The text (probably taken from ocr) which has to be corrected
+     * @return Corrected text
+     */
+    public String correctText(String text){
+
+        //not correcting words with less than minChars characters
+        final int minChars = 3;
```
GiovanniFurlan, line 83: Constant value should be uppercase, also in line 145

app/src/main/java/unipd/se18/ocrcamera/inci/TextAutoCorrection.java
```diff
@@ -0,0 +1,164 @@
...
+            if(match.getDistance() < minDistance) {
+                minDistance = match.getDistance();
+                closest = match.getMatch();
+            }
+        }
+
+        //If no words within maxDistance is found, the same word is returned.
+        return closest.equals("") ? word : closest;
+    }
+}
```
GiovanniFurlan, line 164: Harcoding, could be better to use string.xml in line 32, 66, 102

app/src/main/java/unipd/se18/ocrcamera/inci/TextAutoCorrection.java
```diff
@@ -0,0 +1,164 @@
...
+
+        //find the word with minimum distance
+        int minDistance = Integer.MAX_VALUE;
+        String closest = "";
+        for (BkTreeSearcher.Match<? extends String> match : matches){
+            if(match.getDistance() < minDistance) {
+                minDistance = match.getDistance();
+                closest = match.getMatch();
+            }
+        }
```
GiovanniFurlan, line 159: In this for cycle instead of compare getDistance with minDistance and then calculate new minDistancek, could be better first calculate new minDistance and then compare it. So u dont have to set min Distance to MAX_VALUE before the cicle

app/src/main/java/unipd/se18/ocrcamera/inci/NameMatchIngredientsExtractor.java
```diff
@@ -0,0 +1,96 @@
...
+        //copying list so that sorting doesn't affect original list
+        this.listIngredients = new ArrayList<>(listIngredients);
+
+        //sort by name length so when we search for ingredients in text we match longer names first
+        Collections.sort(this.listIngredients, new Comparator<Ingredient>() {
+            @Override
+            public int compare(Ingredient o1, Ingredient o2) {
+                return o2.getInciName().length() - o1.getInciName().length();
+            }
+        });
```
GiovanniFurlan, line 40: I dont understand why we sort the ingredients by length. Is that essential or gives better performance?

app/src/main/java/unipd/se18/ocrcamera/inci/NameMatchIngredientsExtractor.java
```diff
@@ -0,0 +1,96 @@
...
+        //in mapIndexes we store for each character in the stripped text, the original position
+        int[] mapIndexes = new int[text.length()];
+        StringBuilder strippedTextBuilder = new StringBuilder();
+        for(int i=0; i<text.length(); i++) {
+            char currentChar = text.charAt(i);
+            if(Character.isLetter(currentChar) || Character.isDigit(currentChar)) {
+                mapIndexes[strippedTextBuilder.length()] = i;
+                strippedTextBuilder.append(currentChar);
+            }
+        }
```
GiovanniFurlan, line 65: Why dont u just use String.replace('\W','') instead of map it and use this complex method.
Anyway if it's necessary could be better to create a new method that do that to make the code more linear and to make the code more modular

app/src/main/java/unipd/se18/ocrcamera/inci/NameMatchIngredientsExtractor.java
```diff
@@ -0,0 +1,96 @@
...
+        Collections.sort(foundIngredients, new Comparator<Ingredient>() {
+            @Override
+            public int compare(Ingredient o1, Ingredient o2) {
+                return o1.getStartPositionFound() - o2.getStartPositionFound();
+            }
+        });
+
+        return foundIngredients;
+    }
+}
```
GiovanniFurlan, line 96: Harcoding, could be better to use string.xml in resources. Line 24, 78

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
```diff
@@ -0,0 +1,750 @@
...
+                        nCorrectExtractedIngreds++;
+                        iterator.remove();
+                    }
+                }
+            }
+
+            //make ingredients extraction statistics report
+            int nWrongExtractedIngreds = extractedIngredients.size();
+            float percent = (float)100*nCorrectExtractedIngreds / correctListIngredients.size();
+            if(Float.isNaN(percent)) percent = 0;
```
GiovanniFurlan, line 548: Could u use a better intent for better legibility

app/src/main/java/unipd/se18/ocrcamera/IngredientDetailsFragment.java
```diff
@@ -0,0 +1,171 @@
...
+ * function and an extract from wikipedia.
+ * @author Francesco Pham
+ */
+public class IngredientDetailsFragment extends DialogFragment {
+
+    private static final String TAG = "IngredDetailsFragment";
+
+    private static final String ARG_NAME = "name";
+    private static final String ARG_DESCRIPTION = "description";
+    private static final String ARG_FUNCTION = "function";
```
GiovanniFurlan, line 38: Could be better to use ENUM and string.xml

app/src/main/java/unipd/se18/ocrcamera/IngredientDetailsFragment.java
```diff
@@ -0,0 +1,171 @@
...
+    public IngredientDetailsFragment() {
+        // Required empty public constructor
+    }
+
+    /**
+     * Method for the fragment creation
+     * @param inciName Title to show in the fragment
+     * @param description Description of the ingredient
+     * @param function Functions of the ingredient
+     * @return
```
GiovanniFurlan, line 49: miss return explanation

app/src/main/java/unipd/se18/ocrcamera/IngredientDetailsFragment.java
```diff
@@ -0,0 +1,171 @@
...
+    public void onResume() {
+        super.onResume();
+
+        //set dialog fragment size (width and height values in fragment_ingredients_details.xml do not work)
+        Window window = getDialog().getWindow();
+        if (window != null) {
+            window.setLayout(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.MATCH_PARENT);
+        }
+    }
+}
```
GiovanniFurlan, line 171: Hardcoding at line 115, 142,151

GiovanniFurlan (commented):
Hi Francesco, i've done the review. Mainly i found harcoding, enum missing or  problems in code aspect. 
I also found two or three cases where u could be a better method or i dont understand why u did some actions. Otherwise the code the code is good and i think that using levenstein and three to correct the word is very nice idea to improve performance.

app/src/main/java/unipd/se18/ocrcamera/IngredientDetailsFragment.java
frankplus: For the line number 115 you are right, I will create a method `String buildUrlRequest(String inciname)`, thank you. 
As for number 142 and 151 I think it is a little bit excessive and unnecessary to create a constant for every log message for this reason: these are just log messages for the programmer and it is not visualized in any way by the user, moreover these strings are used just once in the code. It would be easier to put the message directly in the log, this way we can avoid going up and down reading the code. What do you think?

app/src/main/java/unipd/se18/ocrcamera/inci/Inci.java
frankplus: Enums cannot be used in this case because they are just labels and values cannot be set. Here I needed to specify the column numbers in which those values are found in the CSV. Maybe I should have put a comment to be more clear on the functions of those constants, sorry.

app/src/main/java/unipd/se18/ocrcamera/inci/Inci.java
frankplus: I think it is better to separate the try-catch block because even if the reader could not skip the first line, it is preferable to  try continue anyway with reading the CSV file. If you put everything in a try-catch block and the reader throws an exception at `csvReader.skip(1);` then it doesn't even try to read the CSV making all the app completely unfunctional.

app/src/main/java/unipd/se18/ocrcamera/inci/Inci.java
frankplus: I think it is a little bit excessive and unnecessary to create a constant for every log message for this reason: these are just log messages for the programmer and it is not visualized in any way by the user, moreover these strings are used just once in the code. It would be easier to put the message directly in the log, this way we can avoid going up and down reading the code.

app/src/main/java/unipd/se18/ocrcamera/inci/NameMatchIngredientsExtractor.java
frankplus: we sort by length because we want to match longer names first. So for example in the string "SODIUM ACRYLATES COPOLYMER" we don't want to just match "SODIUM" (which also is an ingredient by its own) but the entire ingredient.

app/src/main/java/unipd/se18/ocrcamera/inci/NameMatchIngredientsExtractor.java
frankplus: Because I need to map the indexes between the original text and the text without symbols (because later I need the position of the ingredient name in the original text). If I did `String.replace('\W','')` I would loose any of these informations. 
I didn't think it was much code to create a seperate method but if you think it would be more clear than I will do it, keep in mind that this new method would need to return two things that are needed in the rest of the code: mapIndexes and the strippedText. So If I had to make a separate method then I would have to add a class variable (because two values cannot be returned at once)

app/src/main/java/unipd/se18/ocrcamera/inci/NameMatchIngredientsExtractor.java
frankplus: I don't think we should use string.xml for some logs. In my opinion string.xml should be used only for strings that are printed on screen and its utility is to facilitate the translation of the texts into different languages for example.

app/src/main/java/unipd/se18/ocrcamera/inci/TextAutoCorrection.java
frankplus: Uppercase name are just for class constants that are declared as `static final`. This is just final variable inside a method.
> The names of variables declared **class constants** and of ANSI constants should be all uppercase with words separated by underscores ("_"). (ANSI constants should be avoided, for ease of debugging.)

from https://softwareengineering.stackexchange.com/questions/252243/naming-convention-final-fields-not-static

app/src/main/java/unipd/se18/ocrcamera/IngredientDetailsFragment.java
frankplus: Enums cannot be used in this case because they are just labels without values, in this case I specifically needed some Strings. I cannot find the utility of using string.xml in this case, I think string.xml should be used only for strings that are printed on screen and its utility is to facilitate the translation of the texts into different languages for example. These constants are just for passing parameters from an activity to a fragment.

app/src/main/java/unipd/se18/ocrcamera/IngredientDetailsFragment.java
frankplus: I forgot that sorry, I will add it immediately.

app/src/main/java/unipd/se18/ocrcamera/inci/TextAutoCorrection.java
frankplus: I didn't quite understand sorry, could you be more specific?

app/src/main/java/unipd/se18/ocrcamera/inci/TextAutoCorrection.java
frankplus: I think string.xml should be used only for strings that are printed on screen and its utility is to facilitate the translation of the texts into different languages for example. I think that log texts shoudn't be put into string.xml

app/src/main/java/unipd/se18/ocrcamera/PhotoTester.java
frankplus: Sorry, I didn't quite understand, what do you mean by "intent"?

frankplus:
Hi Giovanni, I really appreciate your review, thanks a lot for your precious time. Sorry for having given to you so much code to be reviewed but I think that this is an important part of the application therefore a thorough examination of the code like this was needed.

PR #118, opened 2019-01-10
Codereview #2 of 4 (Pratesi) from 12 (Moroldo)

leopra:
Hello @lucamoroz 
Here is my code for the review, in summary i was working on the blur recognition so i created a gallery that show all the images from the OCR folder and shows them while calculating the blur value of every image.
I created a new object called BlurObject that holds every image with it's value and i added a specific Adapter to show all the images in a list.
(in the class CameraActivity is only shown the method blurValue that is used, no review needed)
Thanks a lot for your time
Leonardo


app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        setContentView(R.layout.listllayout);
+        super.onCreate(savedInstanceState);
+
+        File path = new File(PHOTOS_FOLDER);
+        fileNames = path.list();
+        if (fileNames != null) {
+            setGallery();
```
lucamoroz, line 58: Rather than calling a method that populates  arrayBlur, you can turn arrayBlur into a local variable which is instantiated by setGallery (that would return an ArrayList). You should also rename "setGallery" with a more descriptive name.

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
```diff
@@ -0,0 +1,143 @@
...
+            }
+        }
+    }
+
+    /**
+     *  Method to show the max value of the blur
+     * @param list list of BlurObject
+     * @return double value
+     */
+    public double findMax(ArrayList<BlurObject> list){
```
lucamoroz, line 110: Java offer the possibility of using Collections.max, so you could avoid re-implementing this method.

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
```diff
@@ -0,0 +1,143 @@
...
+                    arrayBlur.add(obj);
+
+                } catch (FileNotFoundException e) {
+                    e.printStackTrace();
+                    Log.e("err", "filenotfound");
+                } catch (IllegalArgumentException e) {
+                    e.printStackTrace();
+                    Log.e("err", "illegalargument");
+                }
+                count++; //check how many photos load
```
lucamoroz, line 99: If you want to check the size of the arraylist you can directly use blurArray.size() rather than using a counter variable.

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
```diff
@@ -0,0 +1,143 @@
...
+
+    }
+
+    /**
+     * Method that creates a blurObject for every image in the folder
+     *
+     */
+    private void setGallery() {
+        int count =0;
+        for (int i = 0; i < fileNames.length; i++) {                                                          //imagepath.length = number of elements in the folder
```
lucamoroz, line 83: When iterating on a list you can use a "foreach" which makes the code more readable. The syntax in java would be: foreach(String fileName : fileNames) {...}.

Anyway you can directly iterate on each file using File.listFiles() (applied on Photo folder) that return a File[] array.

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
```diff
@@ -0,0 +1,143 @@
...
+            if (getExtension(fileNames[i]) == "jpg") {                                          //checks if it is an image
```
lucamoroz: The '==' compare the reference of an object, you should use String.equals(".."). Anyway by checking only for "jpg" files, you are excluding all the other image extentions like "jpeg" and "png".

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
lucamoroz: When iterating on a list you can use a "foreach" which makes the code more readable. The syntax in java would be: foreach(String fileName : fileNames) {...}.

Anyway you can directly iterate on each file using File.listFiles() (applied on Photo folder) that return a File[] array.

lucamoroz: When iterating on a list you can use a "foreach" which makes the code more readable. The syntax in java would be: foreach(String fileName : fileNames) {...}.

Anyway you can directly iterate on each file using File.listFiles() (applied on Photo folder) that return a File[] array.

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
lucamoroz: The '==' compare the reference of an object, you should use String.equals(".."). Anyway by checking only for "jpg" files, you are excluding all the other image extentions like "jpeg" and "png".

app/src/main/java/unipd/se18/ocrcamera/BlurObject.java
```diff
@@ -0,0 +1,47 @@
...
+    {   this.name=name;
+        this.image= image;
+        this.blur=blur;
+    }
+
+    public BlurObject(String name, Bitmap image)
+    {
+        this.name=name;
+        this.image=image;
+        this.blur= CameraActivity.blurValue(this.image);
```
lucamoroz, line 26: You should document what this method does as I have no access to the documentation

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
lucamoroz: The '==' compare the reference of an object, you should use String.equals(".."). Anyway by checking only for "jpg" files, you are excluding all the other image extentions like "jpeg" and "png".

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
```diff
@@ -0,0 +1,143 @@
...
+
+                try {
+                    File f = new File(PHOTOS_FOLDER, fileNames[i]);                             //creates a new object for each element
+                    Bitmap image = BitmapFactory.decodeStream(new FileInputStream(f));
```
lucamoroz, line 88: The FileInputStream you open is never closed: at the end of your task you should call FileInputStream.close to release the resources to the system.

app/src/main/java/unipd/se18/ocrcamera/BlurObjectAdapter.java
```diff
@@ -0,0 +1,112 @@
...
+        viewHolder.imageView.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+
+                //TODO show images in new Activity with infos
+            }
+        });
+
+        if (item!= null) {
+            Log.e("err", item.toString());
```
lucamoroz, line 81: Why do you log an error if item is not null?

app/src/main/java/unipd/se18/ocrcamera/BlurObjectAdapter.java
```diff
@@ -0,0 +1,112 @@
...
+        viewHolder.position=position;
+        return convertView;
+    }
+
+    /**
+     *  Method to pass the bitmap between activities
+     * @param bitmap
+     * @return
+     */
+    public String createImageFromBitmap(Bitmap bitmap) {
```
lucamoroz, line 94: It's not clear what is the purpose of this method (the spec. is incomplete), anyway you are creating a file with an hardcoded name: you should create a proper variable or resource for it.

app/src/main/java/unipd/se18/ocrcamera/BlurObjectAdapter.java
```diff
@@ -0,0 +1,112 @@
...
+        String fileName = "myImage";
+        try {
+            ByteArrayOutputStream bytes = new ByteArrayOutputStream();
+            bitmap.compress(Bitmap.CompressFormat.JPEG, 100, bytes);
```
lucamoroz, line 98: If you hardcode a value, ('100' in this case), you should document what is the purpose of that value to make your code readable.

app/src/main/java/unipd/se18/ocrcamera/BlurObjectAdapter.java
```diff
@@ -0,0 +1,112 @@
...
+            FileOutputStream fo = getContext().openFileOutput(fileName, Context.MODE_PRIVATE);
+            fo.write(bytes.toByteArray());
+            // close file output
+            fo.close();
+        } catch (Exception e) {
```
lucamoroz, line 103: Why do you catch a generic exception? You should replace it with the proper exception you want to catch (I suppose FileNotFoundException)

lucamoroz (commented):
I've added some inline comments.

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
leopra: I didn't understand completely, can you show me an example? Thanks

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
leopra: In this case i was evaluating the maximum on objects I created, so the max value is evaluated on a specific parameter of the object , i don't really think it would work.

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
leopra: Yes you are right thanks.

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
leopra: Thanks for the observation

app/src/main/java/unipd/se18/ocrcamera/BlurObject.java
leopra: In reality i've added the class cameraActivity where the method is implemented, but yes you are right it could be more clear

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
leopra: I didn't notice, i'm proceeding to modify

app/src/main/java/unipd/se18/ocrcamera/BlurObjectAdapter.java
leopra: It was for troubleshooting, i should remove it

app/src/main/java/unipd/se18/ocrcamera/BlurObjectAdapter.java
leopra: Yes thanks it was a method that i wanted to use later but never used

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
lucamoroz: It would be something like:
(I had to replace '<' with '^' because GitHub doesn't accept these tags)

private List^BlurObject> setGallery(File directory) {
  List^BlurObject> list = new ArrayList^BlurObject>();
  for(File file : directory.listFiles()) {
    //do your stuff
  }
  return list;
}

So inside onCreate you can do:

arrayBlur = setGallery(directory);

Where directory is the file you named "path"

leopra: Ok thanks

PR #119, opened 2019-01-10
Codereview #2 of 4 (Pratesi) from 17 (Ton)

leopra:
Hello @AndreaTonUnipd
Here is my code for the review, in summary i was working on the blur recognition so i created a gallery that show all the images from the OCR folder nad shows them while calculating the blur value of every image.
I created a new object called BlurObject that holds every image with it's value and i added a specific Adapter to show all the images in a list.
(In the class cameraActivity is only shown the method blurValue , no review needed)
Thanks a lot for your time
Leonardo

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
@@ -80,8 +80,8 @@ public class BlurGalleryActivity extends AppCompatActivity {
      */
     private void setGallery() {
         int count =0;
-        for (int i = 0; i < fileNames.length; i++) {                                                          //imagepath.length = number of elements in the folder
-            if (getExtension(fileNames[i]) == "jpg") {                                          //checks if it is an image
+        for (int i = 0; i < fileNames.length; i++) {                                                 //imagepath.length = number of elements in the folder
+            if (getExtension(fileNames[i]).equals("jpg")) {                                          //checks if it is an image
 
                 try {
                     File f = new File(PHOTOS_FOLDER, fileNames[i]);                             //creates a new object for each element
--- a/app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
@@ -1,0 +1,143 @@
+package unipd.se18.ocrcamera;
+
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.os.Bundle;
+import android.os.Environment;
+import android.support.v7.app.AppCompatActivity;
+import android.util.Log;
+import android.widget.ListView;
+import android.widget.TextView;
+import android.widget.Toast;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.util.ArrayList;
+
+import static android.graphics.BitmapFactory.decodeStream;
+
+/** Testing Activity to see if blur could work, shows a gallery with each image associated with it's blur value
+ * Leonardo Pratesi - gruppo 1
+ * Activity accessible by the options menu on the application
+ *
+ */
+public class BlurGalleryActivity extends AppCompatActivity {
+
+    /**
+     *  Directory of the Photos
+     */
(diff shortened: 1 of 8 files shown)


app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
```diff
@@ -0,0 +1,143 @@
...
+
+        File path = new File(PHOTOS_FOLDER);
+        fileNames = path.list();
+        if (fileNames != null) {
+            setGallery();
+        }
+        else {
+            Toast.makeText(this, "No images found", Toast.LENGTH_SHORT).show();
+            this.finish();
+            //Exit from Gallery
```
AndreaTonUnipd, line 63: Could be an idea to show that the gallery is empty and give the option to just get back with the arrow in the top left corner ? 
Instead of just killing the activity and show a Toast.

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
```diff
@@ -0,0 +1,143 @@
...
+                    BlurObject obj = new BlurObject(fileNames[i], image);
+                    arrayBlur.add(obj);
+
+                } catch (FileNotFoundException e) {
+                    e.printStackTrace();
+                    Log.e("err", "filenotfound");
+                } catch (IllegalArgumentException e) {
+                    e.printStackTrace();
+                    Log.e("err", "illegalargument");
+                }
```
AndreaTonUnipd, line 98: Is there any better way to manage this exceptions ?

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
```diff
@@ -0,0 +1,143 @@
...
+
+
+    /**
+     * Method to get file extension leonardo Pratesi
+     * @param file
+     * @return file extension
+     */
+    public String getExtension(String file) {
+        int dotposition = file.lastIndexOf(".");
+        String filename_Without_Ext = file.substring(0, dotposition);
```
AndreaTonUnipd, line 133: Is this in any way useful ? Is it there for possible future implementation ? because I don't see the utility.

app/src/main/java/unipd/se18/ocrcamera/BlurObjectAdapter.java
```diff
@@ -0,0 +1,112 @@
...
+            viewHolder.imageView.setImageBitmap(item.getImage());
+        }
+        viewHolder.position=position;
+        return convertView;
+    }
+
+    /**
+     *  Method to pass the bitmap between activities
+     * @param bitmap
+     * @return
```
AndreaTonUnipd, line 92: @return is missing

app/src/main/java/unipd/se18/ocrcamera/BlurObject.java
```diff
@@ -0,0 +1,47 @@
...
+    public double getBlur()
+    {
+        return blur;
+    }
+
+    public String toString()
+    {
+        return name + " " + Double.toString(blur);
+    }
+
```
AndreaTonUnipd, line 45: missing specs, but I see why you did not write them, is pretty clear anyways, but maybe java doc could be requested.

AndreaTonUnipd (commented):
The code is generally comprehensible, there are some very readable parts and some that are less readable.
Some names are just too generic.

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
leopra: Yes it's a good idea thanks

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
leopra: on the getGallery method I used this to get only the jpg images from the folder

app/src/main/java/unipd/se18/ocrcamera/BlurObjectAdapter.java
leopra: Yes you are right

app/src/main/java/unipd/se18/ocrcamera/BlurGalleryActivity.java
AndreaTonUnipd: Sorry, I can’t find the getGallery method, where is it ?

PR #120, opened 2019-01-11
Codereview #2 of 3 (Porro) from 11 (Cervo)

thomasporro:
Hi,
I've loaded a lot of files, but the methods to review are:
-detectTextRegions and extractTextFromBitmap in ExtractTheText.java
-editBright and isBright in PreProcessing.java
I've put into the pull request also the interface so you can understand more easily the methods but they are not to be revised
Thank You


app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+     * @return the image with the modified brightness
+     * @author Thomas Porro(g1), Giovanni Fasan(g1), Oscar Garrido(g1)
+     */
+    private Bitmap editBright(Bitmap image){
+        /*This variable is used to put a limit to the change of the image's brightness.
+          The value 240 is derived from the fact that in the for loop we try to modify
+          the value of all the pixels of a step, and being the maximum value = 255 (pixel's
+          color maximum value, we put the limit on 240*/
+        final int maxBrightness = 240;
+        final int step = 15;
```
NicoloCervo: Usually constant's names are in UPPER_CASE like IMAGE_TOO_BRIGHT, to make it easier to differentiate them from other variables, it should be in the style sheet

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+
+        //Converts the image into a matrix
+        Mat bright;
```
NicoloCervo: It would be beter to cal this matrix something like imageMatrix, bright doesn't tell what is stored in the variable. modifiedMat at line 249 instead tells you what the variable stores.

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        Mat modifiedMat = new Mat();
+        switch (isBright(bright)) {
+            case IMAGE_TOO_BRIGHT:
+                Log.d(TAG, "Case==IMAGE_TOO_BRIGHT");
+                //Darkens the colour's brightness until it's in an optimal value
+                for(double changeBrightness = 0; changeBrightness != maxBrightness;
+                    changeBrightness -= step){
+                   /*This variable is used to select the type of matrix we want to abtain in the
+                      the convertTo method. If it's negative the type doesn't change*/
+                    int matrixType = -1;
```
NicoloCervo: final int MATRIX_TYPE
same for alpha at line 265
Both could be declared before the switch, no need to declare the same constant 2 times

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+                        //If the conversion failed it returns the original image
+                        try{
+                            return IPUtils.conversionMatToBitmap(modifiedMat);
+                        } catch (ConversionFailedException error){
+                            Log.e(TAG, error.getErrorMessage());
+                            return image;
+                        }
+                    }
+                }
+                break;
```
NicoloCervo: case IMAGE TOO BRIGHT and case IMAGE TOO DARK seem to contain pretty similar code, you could avoid writing it two times by using
```
while(changeBrightness != maxBrightness && != minBrightness){
    if(isBright(bright) == TOO BRIGHT){
        changeBrightness -= step;
    }
   if(isBright(bright) == TOO DARK){
        changeBrightness += step;
  }else { return IPUtils.conversionMatToBitmap(modifiedMat); }
  bright.convertTo(modifiedMat, matrixType, alpha, changeBrightness);
}
```
this is not complete, it's just to get the idea.

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        //Changes the format of the matrix into a RGB one
+        Imgproc.cvtColor(imageMat, brightnessMat, Imgproc.COLOR_RGBA2RGB);
+
+        //Obtain 3 different matrix with the 3 elemental colors
+        List<Mat> color = new ArrayList<>();
+        Core.split(brightnessMat, color);
+
+        /*Each color is multiplied with his luminance.
+          The colors are in order RGB, so to access the che color I use the number 0, 1, 2 in order
+          For more informations see https://en.wikipedia.org/wiki/Relative_luminance*/
```
NicoloCervo, line 174: You could declare 

```
final int RED = 0
final int GREEN = 1
final int GREEN = 1
```
to make the code more readable, but the comment is as good.

maybe add 
//relative luminance = 0.2126R + 0.7152G + 0.0722B 
to the comment since it is pretty relevant to get what you do next but the link is probably enough
You could also save those numbers in constants!

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+     */
+    private BrightnessValue isBright(Mat imageMat){
+        //Converts the image into a matrix
+        Mat brightnessMat = new Mat();
+
+        //Changes the format of the matrix into a RGB one
+        Imgproc.cvtColor(imageMat, brightnessMat, Imgproc.COLOR_RGBA2RGB);
+
+        //Obtain 3 different matrix with the 3 elemental colors
+        List<Mat> color = new ArrayList<>();
```
NicoloCervo: color could be a pretty vague name, maybe something like colorComponent could be easier to comprehend

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        //Sums the matrix of the colors into a single one
+        Mat tempLuminance = new Mat();
+        Mat totalLuminance = new Mat();
+        Core.add(redLuminance , greenLuminance , tempLuminance); //Red + Green = Temp
+        Core.add(tempLuminance , blueLuminance , totalLuminance); //Temp + Blue = Luminance
+
+        //Calculate the sum of the values of all pixels
+        Scalar sum = Core.sumElems(totalLuminance);
+
+        //Determines the image's bit
```
NicoloCervo: not sure what the image's bit is. Is it the number of bits used to represent the color of each pixel?

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        //Calculate the percentage of the brightness
+        double brightness = sum.val[0]/((Math.pow(2,bit)-1)*brightnessMat.rows()
+                *brightnessMat.cols())*2;
+
+        Log.d(TAG, "Brightness:"+brightness);
+
+        /*Bounds to define if the image is dark or bright.
+          The values were decided on the basis of various tests*/
+        double upperBound = 0.9;
+        double lowerBound = 0.4;
```
NicoloCervo: `final double UPPER_BOUND = 0.9;`

app/src/main/java/unipd/se18/ocrcamera/ExtractTheText.java
```diff
@@ -0,0 +1,292 @@
...
+        Log.d(TAG, "Channels = "+rotatedImg.channels());
+        //Crop the resulting image
+        getRectSubPix(rotatedImg, rectSize, rectangle.center, croppedImg);
+        return croppedImg;
+    }
+
+    /**
+     * See DetectTheTextMethods.java.
+     * @author Thomas Porro(g1)
+     */
```
NicoloCervo, line 209: detectTextRegions is easily readable, i don't think it needs changes. good job!

app/src/main/java/unipd/se18/ocrcamera/ExtractTheText.java
```diff
@@ -0,0 +1,292 @@
...
+          Imgproc.getRectSubPix method*/
+        Imgproc.cvtColor(img, img, Imgproc.COLOR_BGRA2BGR);
+
+        /*For each rectangle contained in textContainer extract the rectangle and saves it
+          into a bitmap*/
+        RotatedRect rectangle;
+        while(textContainer.hasNext()){
+            rectangle = (RotatedRect)textContainer.next();
+            Mat croppedMat = crop(rectangle, img);
+            //If the conversion failed return a List with only the original image
```
NicoloCervo: In the specs it sais that if it fails it returns a list with only the full image, but doesn't it return the list of images that it succesfully cropped + the full image? You should find a way to let the methods user know when there is a failure, like writing in the specs to check if a cropped image has the same size of the original(so he knows the method failed)

NicoloCervo (commented):
The code is easy to read and well commented, I just pointed out minor adjustments, good job.

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+            case CV_8U:  bit = 8; break;
+            case CV_8S:  bit = 8; break;
+            case CV_16U: bit = 16; break;
+            case CV_16S: bit = 16; break;
+            case CV_32S: bit = 32; break;
+            case CV_32F: bit = 32; break;
+            case CV_64F: bit = 64; break;
+            default: return BrightnessValue.IMAGE_IS_OK;
+        }
+        //Calculate the percentage of the brightness
```
NicoloCervo: maybe add the formula you use in the comment to make it clearer

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+                    bright.convertTo(modifiedMat, matrixType, alpha, changeBrightness);
+
+                    //Verify if the image is good enough
+                    if(isBright(modifiedMat) == BrightnessValue.IMAGE_IS_OK){
+                        //If the conversion failed it returns the original image
+                        try{
+                            return IPUtils.conversionMatToBitmap(modifiedMat);
+                        } catch (ConversionFailedException error){
+                            Log.e(TAG, error.getErrorMessage());
+                            return image;
```
NicoloCervo: as in the other method(the cropping one) here you have no way of knowing if it failed or not from outside

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        //Call the internal method isBright to detect if the image is bright or dark
+        //and change the brightness according to the number obtained
+        Mat modifiedMat = new Mat();
+        switch (isBright(bright)) {
+            case IMAGE_TOO_BRIGHT:
+                Log.d(TAG, "Case==IMAGE_TOO_BRIGHT");
+                //Darkens the colour's brightness until it's in an optimal value
+                for(double changeBrightness = 0; changeBrightness != maxBrightness;
```
NicoloCervo: THIS COULD CAUSE PROBLEMS! you should end the for if !=-maxBrightness (to be shoure i would put <=, somebody could decide to change step). As it is now it could go into an infinite loop

NicoloCervo (commented):
overall easy to read and well commented, only thing to fix is line 254 of PreProcessing

app/src/main/java/unipd/se18/ocrcamera/ExtractTheText.java
thomasporro: You're right! Maybe  throw an exception and let the user take care of it is a good idea. Let me know if you like it

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
thomasporro: Nice advice

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
thomasporro: Nice advice

app/src/main/java/unipd/se18/ocrcamera/PreProcessing.java
thomasporro: I will change this part in the next commit

thomasporro:
I've uploaded the changes you suggested to me. Tell me if everything is ok

thomasporro:
Btw really nice review!

NicoloCervo:
Ok everything seems ok, good job!

Did you fix the code at 3am? It says your comments are from 12h ago. If you did I must give you props for the commitment.

thomasporro:
Yes ahaha! Because of the exams and my work that was the only time of day when I had time to correct my code.
So i close this pull request. Thank your review it was very helpful

PR #122, opened 2019-01-11
Codereview#2 of 1 (Garrido) from 14 (Romanello)

Nintenking:
Hi Stefano,
here is the code for the review. The methods I'd ask you to look at are:
-"BrightnessValue isBright(Mat imageMat)", 
- "Bitmap editBright(Bitmap image)";
- "boolean isBlurred(Bitmap image)";
All 3 of them are inside the class PreProcessing.java 

I've also left in this pull request the whole ImageProcessing Module, perhaps you can more easily understand the methods this way.
Thank you for your time


imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+     *         IMAGE_IS_BRIGHT if image is too bright,
+     *         IMAGE_IS_DARK if image is too dark.
+     * @author Thomas Porro(g1), Giovanni Fasan(g1), Leonardo Pratesi(g1)
+     */
+    private BrightnessValue isBright(Mat imageMat){
+        //Converts the image into a matrix
+        Mat brightnessMat = new Mat();
+
+        //Changes the format of the matrix into a RGB one
+        Imgproc.cvtColor(imageMat, brightnessMat, Imgproc.COLOR_RGBA2RGB);
```
stefanoromanello: I think you did this for being able to split the color using the next few lines with Core.split right? Shouldn't be better to add this to the comment?

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+     * Detect if the image is bright
+     * @param imageMat the image we want to detect the brightness
+     * @return IMAGE_IS_OK if image is neither too bright nor too dark,
+     *         IMAGE_IS_BRIGHT if image is too bright,
+     *         IMAGE_IS_DARK if image is too dark.
+     * @author Thomas Porro(g1), Giovanni Fasan(g1), Leonardo Pratesi(g1)
+     */
+    private BrightnessValue isBright(Mat imageMat){
+        //Converts the image into a matrix
+        Mat brightnessMat = new Mat();
```
stefanoromanello: I suggest to change the name of this to something like rgbImageMat, in this way I know that I'm working with an RGB image.
Useful also because at line 193 reading the parameter of the switch I know that I'm working with an image

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        int bit;
+        switch ( brightnessMat.depth() ) {
+            case CV_8U:  bit = 8; break;
+            case CV_8S:  bit = 8; break;
+            case CV_16U: bit = 16; break;
+            case CV_16S: bit = 16; break;
+            case CV_32S: bit = 32; break;
+            case CV_32F: bit = 32; break;
+            case CV_64F: bit = 64; break;
+            default: return BrightnessValue.IMAGE_IS_OK;
```
stefanoromanello: What is this value? Aren't these all the possible values for the bits of an image?

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        }
+        //Calculate the percentage of the brightness
```
stefanoromanello: A better description or documentation for what you are doing here should be useful to understand correctly this part of the code.

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+     */
+    private BrightnessValue isBright(Mat imageMat){
+        //Converts the image into a matrix
+        Mat brightnessMat = new Mat();
+
+        //Changes the format of the matrix into a RGB one
+        Imgproc.cvtColor(imageMat, brightnessMat, Imgproc.COLOR_RGBA2RGB);
+
+        //Obtain 3 different matrix with the 3 elemental colors
+        List<Mat> color = new ArrayList<>();
```
stefanoromanello: I suggest to change this to imageColors. In this what I know that I'm working with the image that I want to work with and not with any other object containing colors

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
stefanoromanello: BrightnessValue is an enum and BrightnessValue.IMAGE_IS_OK corresponds to 0. 
The cases contain all the possible bits of an image, right? When I don't get one of them I return IMAGE_IS_OK but if there is an error in the image and opencv is not able to read its depth this is not correct because I can't know if the brightness of the image is really ok.

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+    }
+
+
+    /**
+     * Change the brightness of the image into an optimal one
+     * @param image the image we want to modify the brightness
+     * @return the image with the modified brightness
+     * @author Thomas Porro(g1), Giovanni Fasan(g1), Oscar Garrido(g1)
+     */
+    private Bitmap editBright(Bitmap image){
```
stefanoromanello, line 230: I suggest to change the name of the method to something like "autoAdjustImageBrightness" because I would use "editBright" for methods where I have to tell the new brightness.

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        final int maxBrightness = 240;
+        final int step = 15;
+
+        //Converts the image into a matrix
+        Mat bright;
+        try{
+            bright = IPUtils.conversionBitmapToMat(image);
+        } catch (ConversionFailedException error){
+            Log.e(TAG, error.getErrorMessage());
+            return image;
```
stefanoromanello, line 244: I suggest to return null in this case and manage the exception from where you call editBright() or create a custom exception because is not correct to return an image even if it doesn't have been processed.

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        //Threshold above which the color is out of focus
+        final int threshold = -6118750;
+
+        //Converts the image into a matrix
+        Mat imageMat;
+        try {
+            imageMat = IPUtils.conversionBitmapToMat(image);
+        } catch (ConversionFailedException error){
+            Log.e(TAG, error.getErrorMessage());
+            return false;
```
stefanoromanello: Same comment as line 244

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+                      the formula above.*/
+                    bright.convertTo(modifiedMat, matrixType, alpha, changeBrightness);
+
+                    //Verify if the image is good enough
+                    if(isBright(modifiedMat) == BrightnessValue.IMAGE_IS_OK){
+                        try{
+                            return IPUtils.conversionMatToBitmap(modifiedMat);
+                        } catch (ConversionFailedException error){
+                            Log.e(TAG, error.getErrorMessage());
+                            return image;
```
stefanoromanello: Same comment as line 244

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+                    bright.convertTo(modifiedMat, matrixType, alpha, changeBrightness);
+
+                    //Verify if the image is good enough
+                    if(isBright(modifiedMat) == BrightnessValue.IMAGE_IS_OK){
+                        //If the conversion failed it returns the original image
+                        try{
+                            return IPUtils.conversionMatToBitmap(modifiedMat);
+                        } catch (ConversionFailedException error){
+                            Log.e(TAG, error.getErrorMessage());
+                            return image;
```
stefanoromanello: Same comment as line 244

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+    private Bitmap editBright(Bitmap image){
+        /*This variable is used to put a limit to the change of the image's brightness.
+          The value 240 is derived from the fact that in the for loop we try to modify
+          the value of all the pixels of a step, and being the maximum value = 255 (pixel's
+          color maximum value, we put the limit on 240*/
+        final int maxBrightness = 240;
+        final int step = 15;
+
+        //Converts the image into a matrix
+        Mat bright;
```
stefanoromanello: I suggest to change the name of this object to "imageMap" because as is I can't understand that I'm working with the image

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        //Call the internal method isBright to detect if the image is bright or dark
+        //and change the brightness according to the number obtained
+        Mat modifiedMat = new Mat();
+        switch (isBright(bright)) {
+            case IMAGE_TOO_BRIGHT:
+                Log.d(TAG, "Case==IMAGE_TOO_BRIGHT");
+                //Darkens the colour's brightness until it's in an optimal value
+                for(double changeBrightness = 0; changeBrightness != maxBrightness;
```
stefanoromanello: Maybe I misunderstood but In this way I can end to an infinite loop (avoided by
`if(isBright(modifiedMat) == BrightnessValue.IMAGE_IS_OK)`) , changeBrightness is always different from maxBrightness. I suggest to add a minBrightness=-240 value for example and use 
`for(double changeBrightness = 0; changeBrightness >= minBrightness  ; changeBrightness -= step) `
What do you think?

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+                        }
+                    }
+                }
+                break;
+
+            case IMAGE_TOO_DARK:
+                Log.d(TAG, "Case==IMAGE_TOO_DARK");
+                //The variables are explained in the case above
+                //Lightens the colour's brightness until it's in an optimal value
+                for(double changeBrightness = 0; changeBrightness != maxBrightness;
```
stefanoromanello: Maybe I misunderstood also here but In this way I can end to an infinite loop (avoided also here by
`if(isBright(modifiedMat) == BrightnessValue.IMAGE_IS_OK)`).
I suggest to use
`for(double changeBrightness = 0; changeBrightness <= maxBrightness; changeBrightness += step) `
What do you think?

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+
+    /**
+     * @author Thomas Porro(g1), Oscar Garrido (g1), Giovanni Fasan(g1).
+     * See PreProcessingMethods.java
+     */
+    @Override
+    public boolean isBlurred(Bitmap image) {
+
+        //Total number of color
+        int maxLap = -16777216;
```
stefanoromanello: I suggest to put this as final

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+
+        //Threshold above which the color is out of focus
+        final int threshold = -6118750;
```
stefanoromanello: I suggest to change the name of this to outOfFocusThreshold for a more clear idea of what this is variable is for

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
stefanoromanello: Why returning false if the image has not yet been processed? I suggest to return null in this case and manage the exception from where you call isBlurred() or create a custom exception.

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        Mat laplacianMat8Bit = new Mat();
+        laplacianMat.convertTo(laplacianMat8Bit, CV_8UC1);
+
+        //Create a Bitmap with the given matrix, and obtain all the pixels from it
+        Bitmap laplacianImage;
+        try{
+            laplacianImage = IPUtils.conversionMatToBitmap(laplacianMat8Bit);
+        } catch (ConversionFailedException error){
+            Log.e(TAG, error.getErrorMessage());
+            return false;
```
stefanoromanello: Same as line 336

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
stefanoromanello, line 287: I suggest to add to the comment why you are instantiating this to -16777216

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+
+        //Extracts all the pixels of the laplacian image into the array
+        int[] pixels = IPBuilder.doGetPixels(new IPBuilder.GetPixelsBuilder(laplacianImage)
+                  .withStride(laplacianImage.getWidth())
+                  .withWidth(laplacianImage.getWidth())
+                  .withHeight(laplacianImage.getHeight())
+                  );
+                  
+        //Searches the pixel that has the highest colour range in the RGB format
+        for(int pixel : pixels){
```
stefanoromanello, line 330: I don't understand this part. What are the integer values inside pixels[]?? The brightness of each pixel or a special value obtained using the laplacianImage? At the end of the for cycle maxLap what contains? The brightest pixel?

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+
+        /*Each color is multiplied with his luminance.
+          The colors are in order RGB, so to access the che color I use the number 0, 1, 2 in order
+          For more informations see https://en.wikipedia.org/wiki/Relative_luminance*/
+        Mat redLuminance = new Mat();
+        Core.multiply(color.get(0), new Scalar(0.2126), redLuminance);
+        Mat greenLuminance= new Mat();
+        Core.multiply(color.get(1), new Scalar(0.7152), greenLuminance);
+        Mat blueLuminance = new Mat();
+        Core.multiply(color.get(2), new Scalar(0.0722), blueLuminance);
```
stefanoromanello: I suggest to create constant variables for these values

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
Nintenking: We read in various articles that a way to estimate the sharpness of an image is to apply a Laplace filter and then pick the maximum value. I'll add this information as a comment in the code

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
Nintenking: We created an enum to manage the different situations

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
Nintenking: We understood that we'll always have 8 bit images so we are removing the switch

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
Nintenking: We are making the switch smaller and avoiding the for cycles

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
Nintenking: I thought to create a different object that contains both the image and the TAG, saying if the image was analyzed or not

Nintenking:
Should be ok now, if you think it is correct could you close the pull request so we can merge the module in the Demo?

PR #123, opened 2019-01-11
Codereview#2 of 1 (Garrido) from 8 (Rossi)

Nintenking:
Hi Leonardo,
here is the code for the review. The methods I'd ask you to look at are:
-"BrightnessValue isBright(Mat imageMat)", 
- "Bitmap editBright(Bitmap image)";
- "boolean isBlurred(Bitmap image)";
All 3 of them are inside the class PreProcessing.java 

I've also left in this pull request the whole ImageProcessing Module, perhaps you can more easily understand the methods this way.
Thank you for your time


imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+    /**
+     * Detect if the image is bright
+     * @param imageMat the image we want to detect the brightness
+     * @return IMAGE_IS_OK if image is neither too bright nor too dark,
+     *         IMAGE_IS_BRIGHT if image is too bright,
+     *         IMAGE_IS_DARK if image is too dark.
+     * @author Thomas Porro(g1), Giovanni Fasan(g1), Leonardo Pratesi(g1)
+     */
+    private BrightnessValue isBright(Mat imageMat){
+        //Converts the image into a matrix
```
leopardo-rossi-developer: This comment isn't coherent with the line below; indeed you're not converting the image into a matrix but you're creating a new brightness one.

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        Mat brightnessMat = new Mat();
+
+        //Changes the format of the matrix into a RGB one
+        Imgproc.cvtColor(imageMat, brightnessMat, Imgproc.COLOR_RGBA2RGB);
+
+        //Obtain 3 different matrix with the 3 elemental colors
+        List<Mat> color = new ArrayList<>();
```
leopardo-rossi-developer: If you named this variable "colors" it would have been more coherent with its role

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+
+        /*Each color is multiplied with his luminance.
+          The colors are in order RGB, so to access the che color I use the number 0, 1, 2 in order
+          For more informations see https://en.wikipedia.org/wiki/Relative_luminance*/
+        Mat redLuminance = new Mat();
+        Core.multiply(color.get(0), new Scalar(0.2126), redLuminance);
+        Mat greenLuminance= new Mat();
+        Core.multiply(color.get(1), new Scalar(0.7152), greenLuminance);
+        Mat blueLuminance = new Mat();
+        Core.multiply(color.get(2), new Scalar(0.0722), blueLuminance);
```
leopardo-rossi-developer: All the values you put to initialize a Scalar object can be put into a constant so that you avoid hardcoding

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        Mat tempLuminance = new Mat();
+        Mat totalLuminance = new Mat();
+        Core.add(redLuminance , greenLuminance , tempLuminance); //Red + Green = Temp
+        Core.add(tempLuminance , blueLuminance , totalLuminance); //Temp + Blue = Luminance
+
+        //Calculate the sum of the values of all pixels
+        Scalar sum = Core.sumElems(totalLuminance);
+
+        //Determines the image's bit
+        int bit;
```
leopardo-rossi-developer: What is the purpose of this variable? From the code below I can understand that you use it to calculate image's brightness, but what is its function? If I were you, I've put a more explicative comment to describe it.

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+          The values were decided on the basis of various tests*/
+        double upperBound = 0.9;
+        double lowerBound = 0.4;
+
+        if (brightness > upperBound){             //Image is too bright
+            return BrightnessValue.IMAGE_TOO_BRIGHT;
+        } else if (brightness < lowerBound){        //Image is too dark
+            return BrightnessValue.IMAGE_TOO_DARK;
+        } else {      //Image is neither too bright nor too dark
+            return BrightnessValue.IMAGE_IS_OK;
```
leopardo-rossi-developer, line 219: This value is returned twice: once in switch's default case and here. Why don't you return it only once so that the code is less ambiguous?

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        //Calculate the percentage of the brightness
+        double brightness = sum.val[0]/((Math.pow(2,bit)-1)*brightnessMat.rows()
+                *brightnessMat.cols())*2;
+
+        Log.d(TAG, "Brightness:"+brightness);
+
+        /*Bounds to define if the image is dark or bright.
+          The values were decided on the basis of various tests*/
+        double upperBound = 0.9;
+        double lowerBound = 0.4;
```
leopardo-rossi-developer: I suggest you to transform these two variables into constants so that their value can't be accidentally modified during code execution.

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+     * @return the image with the modified brightness
+     * @author Thomas Porro(g1), Giovanni Fasan(g1), Oscar Garrido(g1)
+     */
+    private Bitmap editBright(Bitmap image){
+        /*This variable is used to put a limit to the change of the image's brightness.
+          The value 240 is derived from the fact that in the for loop we try to modify
+          the value of all the pixels of a step, and being the maximum value = 255 (pixel's
+          color maximum value, we put the limit on 240*/
+        final int maxBrightness = 240;
+        final int step = 15;
```
leopardo-rossi-developer: Remember that constants are always declared uppercased

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+
+        //Converts the image into a matrix
+        Mat bright;
```
leopardo-rossi-developer: This name isn't coherent with the variable's purpose, indeed who reads the code thinks that the matrix you're defining will contain brightness information but it is a conversion of the image passed as parameter instead. For example a possible name could be "imageMatrix"

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+
+        //Call the internal method isBright to detect if the image is bright or dark
+        //and change the brightness according to the number obtained
+        Mat modifiedMat = new Mat();
+        switch (isBright(bright)) {
+            case IMAGE_TOO_BRIGHT:
+                Log.d(TAG, "Case==IMAGE_TOO_BRIGHT");
+                //Darkens the colour's brightness until it's in an optimal value
+                for(double changeBrightness = 0; changeBrightness != maxBrightness;
+                    changeBrightness -= step){
```
leopardo-rossi-developer: Be careful because the for condition will never be satisfied: maxBrightness is grater than zero, changeBrightness starts form zero and at each for iteration it decreases by fifteen so it will be negative and negative and so it won't never be equal to maxBrightness. This isn't a problem if you assume you are always able to correct an image's brightness, but if unfortunately you can't the app will stop in this loop because no return statement is executed

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+                   /*This variable is used to select the type of matrix we want to abtain in the
+                      the convertTo method. If it's negative the type doesn't change*/
+                    int matrixType = -1;
+
+                    /*This variable is used to change the contrast of the matrix, but we want
+                      only modify the brightness so we put the value 1 because the method use
+                      this formula from the documentation:
+                      m(x,y) = saturate _ cast<rType>(alpha(*this)(x,y) + beta)
+                      We called beta as changeBrightness*/
+                    int alpha = 1;
```
leopardo-rossi-developer: If I were you, I've declared this variable and matrixType as constants. This choice makes your code stronger.

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+                Log.d(TAG, "Case==IMAGE_TOO_DARK");
+                //The variables are explained in the case above
+                //Lightens the colour's brightness until it's in an optimal value
+                for(double changeBrightness = 0; changeBrightness != maxBrightness;
+                    changeBrightness += step){
+
+                    //Converts an array to another data type with optional scaling. The variables's
+                    // values are explained in case 1
+                    int matrixType = -1;
+                    int alpha = 1;
```
leopardo-rossi-developer: Same consideration as before for these two variables

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+                    }
+                }
+                break;
+
+            case IMAGE_IS_OK: //Image is neither too bright nor too dark
+                Log.d(TAG, "Case==IMAGE_IS_OK");
+                return image;
+        }
+        return image;
+    }
```
leopardo-rossi-developer, line 276: In general I have two considerations about this method:
1- If switch's cases are consistent of code, I suggest you to use if-else blocks so that in this way code can be a little bit more readable and understandable.
2- In each switch case, if an error occurs while converting the image's matrix into bitmap the original image is returned. Are you really sure this is the better solution? How can you detect that an error has happened outside this class? I suggest you tu return a specific value in case of error during the conversion.

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+    public boolean isBlurred(Bitmap image) {
+
+        //Total number of color
+        int maxLap = -16777216;
+
+        //Threshold above which the color is out of focus
+        final int threshold = -6118750;
+
+        //Converts the image into a matrix
+        Mat imageMat;
```
leopardo-rossi-developer, line 293: Good, this time you chose the right name for this variable!

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+        try {
+            imageMat = IPUtils.conversionBitmapToMat(image);
+        } catch (ConversionFailedException error){
+            Log.e(TAG, error.getErrorMessage());
+            return false;
```
leopardo-rossi-developer: This return like the other one at line 357 it's not correct. Indeed if an error occurred while converting something into a matrix or a bitmap the method's result is false and this means "the given image isn't blurred" when this could not be true. If I where you, I've returned a different value or do something different to distinguish if something went wrong or if the image is really not blurred.

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+            case CV_16U: bit = 16; break;
+            case CV_16S: bit = 16; break;
+            case CV_32S: bit = 32; break;
+            case CV_32F: bit = 32; break;
+            case CV_64F: bit = 64; break;
+            default: return BrightnessValue.IMAGE_IS_OK;
+        }
+        //Calculate the percentage of the brightness
+        double brightness = sum.val[0]/((Math.pow(2,bit)-1)*brightnessMat.rows()
+                *brightnessMat.cols())*2;
```
leopardo-rossi-developer: This formula is a little be "magic". I've preferred if there was a more explicative comment that helps who reads the code to understand what is going on

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
```diff
@@ -0,0 +1,394 @@
...
+                  .withHeight(laplacianImage.getHeight())
+                  );
+                  
+        //Searches the pixel that has the highest colour range in the RGB format
+        for(int pixel : pixels){
+            if(pixel > maxLap){
+                maxLap = pixel;
+            }
+        }
+
```
leopardo-rossi-developer, line 335: If I'm not wrong you determine if an image is or isn't blurred looking to a single pixel value. Is this correct? Don't you have to look to a larger group of pixels to make this decision?

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
Nintenking: You're right, this doesn't make sense. I'll just remove it, must have slipped me from some previous writing

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
Nintenking: You're right, I thought explaining it in the comment was enough but I'll add some constants

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
Nintenking: I'll try to write a more explicative comment!

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
Nintenking: I'll try to consider more the use of constants in the future!

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
Nintenking: We read in various articles that a way to estimate the sharpness of an image is to apply a Laplace filter and then pick the maximum value. I'll add this information as a comment in the code

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
Nintenking: We understood that we'll always have 8 bit images so we are removing the switch

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
Nintenking: You are right, in the end we are changing all the structure of the switch

imageprocessing/src/main/java/com/example/imageprocessing/PreProcessing.java
Nintenking: Ok, then instead of boolean I'll do an enum to differentiate the situations.

Nintenking:
Should be ok now, if you think it is correct could you close the pull request so we can merge the module in the Demo?

PR #125, opened 2019-01-11
Codereview#2 of 17 (Ton) from 11 (Cervo)

AndreaTonUnipd:
Hi, I kindly ask you if you can review my code.
It consists in three classes that belong to a module that implements barcode recognition, and a section from ResultActivity modified for this specific purpose.

To Review:
-Barcode.java (**total**)
-BarcodeRecognizer.java (**total**)
-MLKitBarcode.java (**total**)
-ResultActivity.java (**131-186**)

Barcode Recognition is implemented with MLKit Api by Google (https://firebase.google.com/docs/reference/android/com/google/firebase/ml/vision/barcode/package-summary).

Thanks for your time!

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/BarcodeRecognizer.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/BarcodeRecognizer.java
@@ -1,0 +1,29 @@
+package unipd.se18.barcodemodule;
+
+/**
+ * Useful set a type of recognition
+ * It avoids the single point of failure relative to this process
+ * @author Andrea Ton
+ */
+public class BarcodeRecognizer {
+
+    /**
+     * Ids of the different barcode api
+     */
+    public enum API {
+        zxing
+    }
+
+    /**
+     * Provides a Barcode Recognizer
+     * @param type The id of the recognition api requested
+     * @return the object relative to the choosen type
+     */
+    public static Barcode barcodeRecognizer(API type){
+        switch (type){
+            case mlkit: return new MLKitBarcode();
+            case zxing: return new ZXingBarcode();
+            default: return new MLKitBarcode();
+        }
+    }
+}
--- a/app/src/main/java/unipd/se18/ocrcamera/BarcodeRecognizer.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/BarcodeRecognizer.java
@@ -11,6 +11,7 @@ public class BarcodeRecognizer {
      * Ids of the different barcode api
      */
     public enum API {
+        mlkit,
         zxing
     }
 
--- a/app/src/main/java/unipd/se18/ocrcamera/MLKitBarcode.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/MLKitBarcode.java
@@ -1,0 +1,88 @@
+package unipd.se18.barcodemodule;
+
+import android.graphics.Bitmap;
+import android.support.annotation.NonNull;
+import android.util.Log;
+import com.google.android.gms.tasks.OnFailureListener;
+import com.google.android.gms.tasks.OnSuccessListener;
+import com.google.android.gms.tasks.Task;
+import com.google.firebase.ml.vision.FirebaseVision;
+import com.google.firebase.ml.vision.barcode.FirebaseVisionBarcode;
+import com.google.firebase.ml.vision.barcode.FirebaseVisionBarcodeDetector;
+import com.google.firebase.ml.vision.barcode.FirebaseVisionBarcodeDetectorOptions;
+import com.google.firebase.ml.vision.common.FirebaseVisionImage;
+import java.util.List;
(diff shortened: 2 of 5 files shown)


app/src/main/java/unipd/se18/ocrcamera/BarcodeRecognizer.java
```diff
@@ -0,0 +1,29 @@
...
+ * Useful set a type of recognition
+ * It avoids the single point of failure relative to this process
+ * @author Andrea Ton
+ */
+public class BarcodeRecognizer {
+
+    /**
+     * Ids of the different barcode api
+     */
+    public enum API {
```
NicoloCervo: shouldn't you add mlkit here? Otherwise API will never assume the value.

app/src/main/java/unipd/se18/ocrcamera/BarcodeRecognizer.java
```diff
@@ -0,0 +1,29 @@
...
+        zxing
```
NicoloCervo, line 15: From the other class you use BarcodeRecognizer.API.mlkit, does it work if you dont add mlkit to this enum. Also, zxing and mlkit would be constants so they should be alluppercase. (not really sure for enums tho)

app/src/main/java/unipd/se18/ocrcamera/MLKitBarcode.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+public class MLKitBarcode implements Barcode{
+
+    //actual barcode String returned by main method decodeBarcode
+    private String barcode = "";
+    //text of the decoding error
+    private static final String DECODE_ERROR = "ERROR: Barcode decoding unsuccessful, please try again.";
+    //text of the possible exception related to the CountdownLatch
+    private static final String LATCH_ERROR = "Fatal error! Please try again.";
+    //possibility to limit barcode format recognized to improve performance
```
NicoloCervo, line 29: to make this an option you should ask the format in a method, otherwise the only way to change it is by modifying this class. Format could be a parameter of this class's constructor.

app/src/main/java/unipd/se18/ocrcamera/MLKitBarcode.java
```diff
@@ -0,0 +1,88 @@
...
+        //Get the detector, modularized from the main method, so it's possible to make independent changes
+        //method to find and decode a barcode in the image
+        Task<List<FirebaseVisionBarcode>> result = detector.detectInImage(image)
+                //adding a listener for the success result
+                .addOnSuccessListener(new OnSuccessListener<List<FirebaseVisionBarcode>>() {
+                    @Override
+                    //it gets a List of barcodes as parameter, which is the output of the barcode detector
+                    public void onSuccess(List<FirebaseVisionBarcode> barcodes) {
+                        if (!(barcodes.isEmpty())) {
+                            //if not empty get the string of the barcode value, which is in the first position of the List
```
NicoloCervo, line 57: is there a reason to only get the first barcode? if an image contains multiple barcods and at least two are recognized, what happens to the second?

app/src/main/java/unipd/se18/ocrcamera/MLKitBarcode.java
```diff
@@ -0,0 +1,88 @@
...
+                            barcode = barcodes.get(0).getRawValue();
+                        }else{
+                            //if no barcode is found, return empty string
```
NicoloCervo, line 60: Should tell this in the specs too.

NicoloCervo (commented):
Didn't find many changes necessary, the code is comprehensible, I just pointed out some parts that I dont get, probably my fault.

app/src/main/java/unipd/se18/ocrcamera/BarcodeRecognizer.java
AndreaTonUnipd: You are right Nicolò, the mlkit id is missing but it should be there, in the original file it is so I can't explain why it's missing, since the files are just copied-pasted for this pull request :).
Anyway, my bad, I'm gonna add it now.
Thanks!

app/src/main/java/unipd/se18/ocrcamera/MLKitBarcode.java
AndreaTonUnipd: This implementation is done for just one barcode, assuming that a product will have a univocal barcode (in the EAN13 European standard), so it is identified by a single code (retrieved from the barcode). Anyway multiple barcode recognition can be a future development.

app/src/main/java/unipd/se18/ocrcamera/MLKitBarcode.java
AndreaTonUnipd: You are right, in this project stage every barcode recognition API decides for itself how to operate, if I add the format selection outside this class, it must work in the same way for all the apis.

PR #126, opened 2019-01-11
Codereview#2 of 15 (Bedin) from 8 (Rossi)

ghost:
Dear Leonardo,

here the code of four classes  I would like you to review.
Three of them are part of a module that resolves the string of a barcode into the name of a product.
The last one (**ZXingBarcode**) is my contribution to the [Barcode module](https://github.com/PrandiniUniPD/OCRCamera/tree/barcode/barcodeModule) that Andrea Ton made.

If you need explanations, in particular on the external libraries and API used, feel free to comment about here.
Thank you for your work,

Elia


app/src/main/java/unipd/se18/ocrcamera/EANResolve.java
```diff
@@ -0,0 +1,27 @@
...
+ * Useful set a type of recognition
+ * It avoids the single point of failure relative to this process
+ * @author Elia Bedin
+ */
+public class EANResolve {
+    /**
+     * Ids of the different EAN resolver API
+     */
+    public enum API {
+        mignify
```
leopardo-rossi-developer: Enum cases are usually uppercased

app/src/main/java/unipd/se18/ocrcamera/MignifyResolver.java
```diff
@@ -0,0 +1,111 @@
...
+                            MASHAPE_AUTH)
+                    //Accept and Content-Type selection header
+                    .header("Accept", "application/json")
+                    //Type of response we expect
+                    .asJson();
+        } catch (UnirestException e) {
+            Log.e("Unirest error", e.getMessage());
+        }
+        if (response == null) {
+            product = HTTP_ERROR;
```
leopardo-rossi-developer: This if-else is redundant, indeed you could put this line (66) in the catch block before and all the lines inside the else block could have been put below line 61 because if you never enter in the catch clause you can be sure response is different from null. If you choose this solution all the subsequent catch clauses can be refer to the same try block and the code is more readable and understandable.

app/src/main/java/unipd/se18/ocrcamera/MignifyResolver.java
```diff
@@ -0,0 +1,111 @@
...
+                //The server responds with either "ok" or "error"
+                String status = jsonResponse.getString("status");
+                if (status.equals("ok")) {
+                    //If the EAN code was found, in the JSON there is an list of products, each
+                    //one with name, brand and language
+                    JSONArray productList = jsonResponse.getJSONArray("results");
+                    String[] productNames = new String[productList.length()];
+                    //Future proofing usage of brands and language, still unimplemented
+                    String[] productBrands = new String[productList.length()];
+                    String[] productLanguages = new String[productList.length()];
```
leopardo-rossi-developer, line 99: If I were you, I've put all these declarations (from line 78 to line 81) inside the else block below. In this way you are sure there are some products in the JSON response and so all these variables instantiation is more useful.

app/src/main/java/unipd/se18/ocrcamera/MignifyResolver.java
```diff
@@ -0,0 +1,111 @@
...
+                        }
+
+                        //Selecting the first name in the list as the result product name.
+                        //It should be the most plausible product name.
+                        //In the future a better selection will be made, checking brand and language
+                        product = productNames[0];
+                    }
+                } else if(status.equals("error")){
+                    //In case the EAN wasn't valid, the error in the JSON will be returned
+                    product = jsonResponse.getString("message");
```
leopardo-rossi-developer: Are you sure this error message is useful for the user who is looking for the product corresponding to the barcode? Maybe you could assign to product variable a more meaningful message to explain the user what has happened.

app/src/main/java/unipd/se18/ocrcamera/MignifyResolver.java
```diff
@@ -0,0 +1,111 @@
...
+                }
+            } catch (JSONException e){
+                product = JSON_ERROR;
+                Log.e("JSON error", e.getMessage());
+            }
+        }
+        return product;
+    }
+}
```
leopardo-rossi-developer, line 130: In general in this method there's a lot of hardcoding. If I were you, I've stored all JSON fields' keys into some constants or, even better, I've used android string resources to save them. In this way if in the future these names will change, you have to modify them in one place and you're not forced to look through the whole code.

app/src/main/java/unipd/se18/ocrcamera/ZXingBarcode.java
```diff
@@ -0,0 +1,80 @@
...
+
+        //int array needed by the RGBLuminanceSource constructor
+        int[] pixels = new int[bitmap.getWidth()*bitmap.getHeight()];
+        //getPixels copies pixel data from the Bitmap into the 'pixels' int array
+        //The three zeros (that will always be zeros) are, in order:
+        //  offset, the first index to write into pixels[];
+        //  the x coordinate of the first pixel to read from the bitmap;
+        //  the y coordinate of the first pixel to read from the bitmap
+        bitmap.getPixels(pixels, 0, bitmap.getWidth(), 0, 0, bitmap.getWidth(),
+                bitmap.getHeight());
```
leopardo-rossi-developer: Event if the three zeros won't change their value in the future if I were you, I've put them into three constants each one with the corresponding name you wrote in the comment. The code is stronger and if, for some reason the comment will be erased, who reads the code can immediately understand what is going on.

leopardo-rossi-developer (commented):
I've modified a comment written before

app/src/main/java/unipd/se18/ocrcamera/EANResolve.java
ghost: Thank you for pointing this out, I definitely missed that re-checking my code!

app/src/main/java/unipd/se18/ocrcamera/MignifyResolver.java
ghost: Yeah, now I see it, thank you, this will be changed soon.

app/src/main/java/unipd/se18/ocrcamera/MignifyResolver.java
ghost: You are absolutely right, this will be changed soon.

app/src/main/java/unipd/se18/ocrcamera/MignifyResolver.java
ghost: I probably should have explained this part of the code better.
Basically, this is the case where the barcode is invalid, which will never happens since the Barcode Module already check for that.
I checked anyway what the API would respond to a wrong barcode, and it was a good and understandable response.
I do agree I should normalize the response with an error string I decide, but if wasn't urging since this else should never be entered.

app/src/main/java/unipd/se18/ocrcamera/MignifyResolver.java
ghost: Pietro Prandini was kind enough to suggest to use string.xml for this usage.
So I do agree with the both of you, I should move the JSON and fields and error constants to the string file.

app/src/main/java/unipd/se18/ocrcamera/ZXingBarcode.java
ghost: I'm asking you this because you definitely have more experience than me coding.
I didn't wrote this static zeros in constants because I thought they would just make the variable declaration cluttered, since this zeros have really no meaningful information.
If, however, you say that it is better practice anyway, even for this type of costants, I'll change it.

app/src/main/java/unipd/se18/ocrcamera/MignifyResolver.java
ghost: Since this is a Module, and it doesn't have the context, I cannot access the local string.xml, but only the global one with `Resources.getSystem().getString(android.R.string.supercool)`, correct me if I'm wrong.
So for the moment I'll put everything meaningful in constants.

ghost:
Thank you very much for your feedback!
I've updeted the code base on major review aspects.

PR #127, opened 2019-01-11
Codereview#2 of 15 (Bedin) from 12 (Moroldo)

ghost:
Dear Luca,

here the code of four classes  I would like you to review.
Three of them are part of a module that resolves the string of a barcode into the name of a product.
The last one (**ZXingBarcode**) is my contribution to the [Barcode module](https://github.com/PrandiniUniPD/OCRCamera/tree/barcode/barcodeModule) that Andrea Ton made.

If you need explanations, in particular on the external libraries and API used, feel free to comment about here.
Thank you for your work,

Elia


app/src/main/java/unipd/se18/MignifyResolver.java
```diff
@@ -0,0 +1,111 @@
...
+    public String decodeEAN(String EANCode) {
+
+        //Adding to the URL request the EAN code taken from the barcode
+        String fullURL = URL + EANCode;
+        //HTTP response from get method to Mignify API
+        HttpResponse<JsonNode> response = null;
+        //Unirest library useful for making an HTTP request for the Mashape API
+        //More at http://unirest.io/java
+        try {
+            response = Unirest
```
lucamoroz, line 74: Reading the Unirest documentation for java, I saw that this kind of request blocks the application; have you considered doing an asynchronous request?
You can see some examples here, always with Unirest:
https://www.programcreek.com/java-api-examples/index.php?api=com.mashape.unirest.http.async.Callback

app/src/main/java/unipd/se18/ZXingBarcode.java
```diff
@@ -0,0 +1,80 @@
...
+            code = result.getText();
+        }
+        catch (NotFoundException e) {
+            //If no barcode if found, "" is returned
+            Log.e("NoBarcode", "Error getting barcode from bitmap.", e);
+        }
+        catch (ChecksumException | FormatException e) {
+            //If a barcode is found but decoding it leads to an error, a string that report an
+            // error occurred is returned
+            code = DECODE_ERROR;
```
lucamoroz, line 77: Returning an error in the result String can lead to misunderstanding (you are doing the same also in MignifyResolver). You can throw a proper exception that will be caught inside the calling method, so the utilizer can also set a custom error message that would be indipendent from your implementation.

lucamoroz: Alternatively you could create a listener interface that defines the methods (for example) "onBarcodeRecognized(String code)" and "onBarcodeError(int code)", if you are interested you can check the work done by Prandini for the text recognizer here:
https://github.com/PrandiniUniPD/OCRCamera/tree/demo_g1%2Bg2%2Bg3/textrecognizer/src/main/java/unipd/se18/textrecognizer

app/src/main/java/unipd/se18/MignifyResolver.java
```diff
@@ -0,0 +1,111 @@
...
+
+    //The string resolved from an EAN product code
+    private String product = "";
+    //The string passed to the get method, postponed by the EAN string code
+    private static final String URL =
+            "https://mignify.p.mashape.com/gtins/v1.0/productsToGtin?gtin=";
+    //The authentication string you get from
+    //https://market.mashape.com/mignify/product-intelligence#
+    //after you subscribe to the Mignify API
+    private static final String MASHAPE_AUTH = "Nx432kOz2Wmsh83IAjUv6cUaxT7Rp1tDUbKjsnu7ueSdRFdOPK";
```
lucamoroz, line 32: Shouldn't this auth key be hidden?

lucamoroz (commented):
I've added some inline comments. Maybe the decodeEAN method could be splitted into smaller methods to make it more "light" to read.
Overall I found the code well documented.

app/src/main/java/unipd/se18/MignifyResolver.java
ghost: Yeah, I considered it.
For the first implementation I did the basic not async request.
But I do agree with you, this will be absolutely changed in the future, thank you!

app/src/main/java/unipd/se18/ZXingBarcode.java
ghost: Sorry about this.
Since I've basically always have written C programs, I'm not that used to proper exception handling and error return.
So I have to thank you for pointing me to the proper way, I'll study both your alternatives.
If I were you I wont expect some changes in the near future, I don't want to rush stuff done without properly studing how it should be done.

app/src/main/java/unipd/se18/MignifyResolver.java
ghost: I did some reseach about this.
What I found is, you can't 100% hide the value of a literals string.
To the user, yes.
To people decompiling you java files, no, even if you use libraries that obscure you code.
The [way I found](https://medium.com/@anujjindal7/android-string-literals-obfuscation-in-code-6700a85d5cd1) works the best is to write this constants in a C _library_, build it with [ndk](https://developer.android.com/ndk/guides/ndk-build) and you get a .so which isn't reversible.
Since this isn't that big of a deal that at the moment the authenthication key is visible, I'll leave this like this.
But I will probably at least hide it to the user in the future.

ghost:
Thank you very much for your feedback!
I've updeted the code base on major review aspects.

lucamoroz:
I saw you have also created some variables to store the keys used to parse the JSON response. Thank you, I heedlessy forgot to mention that.

PR #128, opened 2019-01-12
Codereview#2 of 10 (Balzan) from 3 (Porro)

BalzPit:
I added a new Activity "MainAllergensActivity" that lets the user select new Allergens from a list to let him/her be notified if the selected allergens are among the ingredients found on a given cosmetic product.
The activity is made of 2 fragments so that the user can swipe between them, one to search new Allergens to add, and "UserAllergensFragment" ( the one i included for the review) that only shows the User's list.
An allergen is showed with its CommonName (to help the user find it more easily) and a button next to it that, if pressed, allows the user to select said Allergen.
The lists of allergens in the activity are adapted through the custom adapter "AllergensListAdapter".

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
@@ -1,0 +1,103 @@
+package unipd.se18.ocrcamera;
+
+import android.content.Context;
+import android.support.annotation.NonNull;
+import android.util.Log;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.ArrayAdapter;
+import android.widget.CompoundButton;
+import android.widget.TextView;
+import android.widget.ToggleButton;
+
+import java.util.ArrayList;
+
+/*
+ * This class is used to initiate the listViews in MainAllergensActivity
+ * Also defined here are the instructions performed when a user selects a new allergen from the list
+ * @author Pietro Balzan
+ */
+public class AllergenListAdapter extends ArrayAdapter<Allergen> {
+
+    private static final String TAG = "AllergenListAdapter";
+    private Context mContext;
+    private int mResource;
+    private AllergensManager mAllergensManager;
+    private ArrayList<Allergen> selectedAllergens;
+
+    /**
+     * Default constructor for an AllergensListAdapter
+     * @param context of the calling activity
+     * @param resource int of the resource layout file of the view to adapt
+     * @param objects an arraylist of objects to be adapted
+     */
+    AllergenListAdapter(Context context, int resource, ArrayList<Allergen> objects) {
+        //adapt the Allergens List "objects"
+        super(context, resource, objects);
+        mContext = context;
+        //in our practical case we will be using allergen_single.xml
+        mResource = resource;
+        //set an AllergenManager used to modify users' allergens list
+        mAllergensManager= new AllergensManager(mContext);
+        selectedAllergens = mAllergensManager.getSelectedAllergensList();
+    }
+
+    /**
+     * this method deals with the adaptation of the single allergen view
+     * to the ListView of AllergensSearchFragment
+     * @param position of the Allergen
+     * @param convertView the view to be adapted
+     * @param parent viewGroup
+     * @return convertView the adapted View
+     */
+    @NonNull
+    @Override
+    public View getView(final int position, View convertView, ViewGroup parent) {
+        try {
+            //Allergen information
+            final Allergen mAllergen= getItem(position);
+            String name = mAllergen.getCommonName();
+            boolean selected = selectedAllergens.contains(mAllergen);
+
+            //create the Allergen in the layout
+            LayoutInflater inflater = LayoutInflater.from(mContext);
+            convertView = inflater.inflate(mResource, parent, false);
+
+            //set TextView to name of the Allergen
+            Log.i(TAG, "Set allergen name to "+name);
+            TextView allergenTv = convertView.findViewById(R.id.singleAllergenTv);
+            allergenTv.setText(name);
+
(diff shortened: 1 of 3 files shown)


app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+     * @param convertView the view to be adapted
+     * @param parent viewGroup
+     * @return convertView the adapted View
+     */
+    @NonNull
+    @Override
+    public View getView(final int position, View convertView, ViewGroup parent) {
+        try {
+            //Allergen information
+            final Allergen mAllergen= getItem(position);
```
thomasporro, line 59: In these files the only getItem method I saw is in MainAllergensActivity but it returns a Fragment object. So this getItem(position) is into another class?

app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+            String name = mAllergen.getCommonName();
+            boolean selected = selectedAllergens.contains(mAllergen);
+
+            //create the Allergen in the layout
+            LayoutInflater inflater = LayoutInflater.from(mContext);
+            convertView = inflater.inflate(mResource, parent, false);
```
thomasporro, line 65: Maybe you sould explain why you decided not to consider inflater as a "child" of parent. You could explain it while creating a boolean variable before the method. In this way you'll explain better the method and you'll avoid hardcoding

app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+                    else {
+                        Log.i(TAG, "add allergen to user's list");
+                        selectedAllergens.add(mAllergen);
+                    }
+                    Log.i(TAG, "Lista degli allergeni: "+selectedAllergens.toString());
+                    mAllergensManager.updateSelectedAllergens(selectedAllergens);
+                }
+            });
+        }
+        catch (NullPointerException npe){
```
thomasporro, line 97: Is it possible the use a smaller try-catch block?

app/src/main/java/unipd/se18/ocrcamera/UserAllergensFragment.java
```diff
@@ -0,0 +1,47 @@
...
+     * @param container
+     * @param savedInstanceState
+     * @return fragmentView
+     * @author Pietro Balzan
+     */
+    @Nullable
+    @Override
+    public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState) {
+        //use the layout file of this fragment to be inflated intoa view
+        View fragmentView= inflater.inflate(R.layout.fragment_users_allergens, container,  false);
```
thomasporro, line 35: The same thing I've said before

app/src/main/java/unipd/se18/ocrcamera/UserAllergensFragment.java
```diff
@@ -0,0 +1,47 @@
...
+        allergensView = (ListView) fragmentView.findViewById(R.id.users_allergens_list_view);
```
thomasporro, line 36: Here and in the following lines i've understand that you setted up the adapter, but I can't find out where you extract/use it. Can you explain it to me?

app/src/main/java/unipd/se18/ocrcamera/MainAllergensActivity.java
```diff
@@ -0,0 +1,100 @@
...
+        public Fragment getItem(int position) {
+            return mFragmentList.get(position);
+        }
+
+        @Override
+        public int getCount() {
+            return mFragmentList.size();
+        }
+    }
+}
```
thomasporro, line 100: This class seems to be ok. Nice job!

thomasporro (commented):
The code is easy to read. There are things that are not entirely clear to me so I asked you in the comments to explain to me

app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
BalzPit: I think what happens here is that since we are overriding getView the program automatically uses the list of Allergens ("objects") that was passedto the "  super(context, resource, objects);" at line 37. So i never wrote a "getItem(int i)" method and the one you see is just a standard method that works on ArrayLists of any object.

app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
BalzPit: good suggestion

app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
BalzPit: Yes, i forgot to write in the pull request summary that since I am still inexperienced I didn't really know how to handle this NullPointerException catch, but I did know this was probably not the way to do it. Since there are a few lines that could cause that exception I decided to make only one big try.

app/src/main/java/unipd/se18/ocrcamera/UserAllergensFragment.java
BalzPit: So, when I initialize mAllergensListAdapter at line 42 it creates an adapter for a List (userList in this case) with all the information it needs to create the layout we want, and it is later used to adapt the ListView "allergensView" at line 43.
-> allergensView.setAdapter(**mAllergensListAdapter**);

BalzPit:
Thanks for spending your time reading my code to make this review, it was very helpful!

PR #129, opened 2019-01-12
Codereview#2 of 10 (Balzan) from 17 (Ton)

BalzPit:
I added a new Activity "MainAllergensActivity" that lets the user select new Allergens from a list to let him/her be notified if the selected allergens are among the ingredients found on a given cosmetic product.
The activity is made of 2 fragments so that the user can swipe between them, one to search new Allergens to add, and "UserAllergensFragment" ( the one i included for the review) that only shows the User's list.
An allergen is showed with its CommonName (to help the user find it more easily) and a button next to it that, if pressed, allows the user to select said Allergen.
The lists of allergens in the activity are adapted through the custom adapter "AllergensListAdapter".

### DIFF
--- a/app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
+++ b/app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
@@ -1,0 +1,103 @@
+package unipd.se18.ocrcamera;
+
+import android.content.Context;
+import android.support.annotation.NonNull;
+import android.util.Log;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.ArrayAdapter;
+import android.widget.CompoundButton;
+import android.widget.TextView;
+import android.widget.ToggleButton;
+
+import java.util.ArrayList;
+
+/*
+ * This class is used to initiate the listViews in MainAllergensActivity
+ * Also defined here are the instructions performed when a user selects a new allergen from the list
+ * @author Pietro Balzan
+ */
+public class AllergenListAdapter extends ArrayAdapter<Allergen> {
+
+    private static final String TAG = "AllergenListAdapter";
+    private Context mContext;
+    private int mResource;
+    private AllergensManager mAllergensManager;
+    private ArrayList<Allergen> selectedAllergens;
+
+    /**
+     * Default constructor for an AllergensListAdapter
+     * @param context of the calling activity
+     * @param resource int of the resource layout file of the view to adapt
+     * @param objects an arraylist of objects to be adapted
+     */
+    AllergenListAdapter(Context context, int resource, ArrayList<Allergen> objects) {
+        //adapt the Allergens List "objects"
+        super(context, resource, objects);
+        mContext = context;
+        //in our practical case we will be using allergen_single.xml
+        mResource = resource;
+        //set an AllergenManager used to modify users' allergens list
+        mAllergensManager= new AllergensManager(mContext);
+        selectedAllergens = mAllergensManager.getSelectedAllergensList();
+    }
+
+    /**
+     * this method deals with the adaptation of the single allergen view
+     * to the ListView of AllergensSearchFragment
+     * @param position of the Allergen
+     * @param convertView the view to be adapted
+     * @param parent viewGroup
+     * @return convertView the adapted View
+     */
+    @NonNull
+    @Override
+    public View getView(final int position, View convertView, ViewGroup parent) {
+        try {
+            //Allergen information
+            final Allergen mAllergen= getItem(position);
+            String name = mAllergen.getCommonName();
+            boolean selected = selectedAllergens.contains(mAllergen);
+
+            //create the Allergen in the layout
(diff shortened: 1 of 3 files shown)


app/src/main/java/unipd/se18/ocrcamera/UserAllergensFragment.java
```diff
@@ -0,0 +1,47 @@
...
+     * @return fragmentView
+     * @author Pietro Balzan
+     */
+    @Nullable
+    @Override
+    public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState) {
+        //use the layout file of this fragment to be inflated intoa view
+        View fragmentView= inflater.inflate(R.layout.fragment_users_allergens, container,  false);
+        allergensView = (ListView) fragmentView.findViewById(R.id.users_allergens_list_view);
+
```
AndreaTonUnipd, line 37: Shouldn't you move these procedures, from here to .setAdapter(), to the OnViewCreated() method ? 
As I remember it is unsafe to call getActivity() in OnCreateView(), in this method you should only create fragment's view object hierarchy.

app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+     * this method deals with the adaptation of the single allergen view
+     * to the ListView of AllergensSearchFragment
+     * @param position of the Allergen
+     * @param convertView the view to be adapted
+     * @param parent viewGroup
+     * @return convertView the adapted View
+     */
+    @NonNull
+    @Override
+    public View getView(final int position, View convertView, ViewGroup parent) {
```
AndreaTonUnipd, line 56: I was wondering if this method could have been a little more modular, maybe to be easier for further unit testing, since the more dependencies it has, the more complicated to test it becomes.
Anyway this is the only valid reason I see to report modularity issue in this method, otherwise it looks fine to me.

app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+                    }
+                    Log.i(TAG, "Lista degli allergeni: "+selectedAllergens.toString());
+                    mAllergensManager.updateSelectedAllergens(selectedAllergens);
+                }
+            });
+        }
+        catch (NullPointerException npe){
+            Log.e(TAG, "Missing Allergens' parameters");
+            npe.printStackTrace();
+        }
```
AndreaTonUnipd, line 100: Is there any better way to manage this exception ? What happens if the exception happens ?

app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+        try {
+            //Allergen information
+            final Allergen mAllergen= getItem(position);
+            String name = mAllergen.getCommonName();
+            boolean selected = selectedAllergens.contains(mAllergen);
+
+            //create the Allergen in the layout
+            LayoutInflater inflater = LayoutInflater.from(mContext);
+            convertView = inflater.inflate(mResource, parent, false);
+
```
AndreaTonUnipd, line 66: what does "false" stand for in this method ?

AndreaTonUnipd (commented):
Overall the code is clear, I can't find any real big issue.
I just found the variables name a bit confusing to read, with the only "m" added in front of the names, but it's fine after you get used to read it.
Also, by reading just this classes without some others, some method could need a little more explanation, since their class is not provided and so there are no specs about them.

app/src/main/java/unipd/se18/ocrcamera/UserAllergensFragment.java
BalzPit: Yes,thank you for the suggestion. I just made a review for a code that used onViewCreated as you said. Being still relatively new to Android, I didn't know that i should have done that, my googling skills probably weren't enough this time.

app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
BalzPit: It means that i am not considering inflater as a "child" of "parent". I should probably make a boolean variable before to explain better.

app/src/main/java/unipd/se18/ocrcamera/AllergenListAdapter.java
BalzPit: As i said to the other reviewer, i forgot to write in the pull request summary that since I am still inexperienced I didn't really know how to handle this NullPointerException, but I did know this was probably not the way to do it. I thought about giving a message to the user or even closing the Activity, but both seemed too User-unfriendly to me because there is nothing he can do to solve the problem.

BalzPit:
Thanks for spending your time reading my code to make this review, it was very helpful!

PR #130, opened 2019-01-12
CodeReview#2 of 5 (bullaku) from 10 (Balzan)

TaulantBullaku:
hello, I kindly ask you if you can do the codereview of my code inside ForumLogin.java


forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
```diff
@@ -0,0 +1,174 @@
...
+    private final String KEY_USERNAME = "username";
+
+    private EditText usernameEditText;
+    private EditText pwdEditText;
+    private Button loginButton;
+
+    private String userName;
+    private String userPwd;
+
+    private Login_VM viewModel;
```
BalzPit, line 68: The global variables are very well commented and easy to understand, except this one (viewModel). I had to look for its class in the other files to know what it was for.

forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
```diff
@@ -0,0 +1,174 @@
...
+             * user an error message, that is performed through the use of a toast
+             *
+             * @param message The error message about what was wrong with the login request
+             * @author Alberto Valente (g2), Taulant Bullaku (g2)
+             */
+            @Override
+            public void onLoginFailure(String message) {
+
+                if (message != null) {
+                    Toast.makeText(getContext(), message, Toast.LENGTH_LONG).show();
```
BalzPit, line 157: Here I can't really understand where the content of the String "message" comes from. Is it something in the "Login_VM" class? I didn't seem to find anything there though.

forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
```diff
@@ -0,0 +1,174 @@
...
+                userPwd = pwdEditText.getText().toString();
+
+                //Then the credentials are handed to the view model method to be checked
+                viewModel.loginToForum(getContext(), userName, userPwd);
+            }
+        });
+
+    }
+
+    public interface OnFragmentInteractionListener
```
BalzPit: I see that you initialized "mListener" at line 36 but it does not do anything. Does that TODO mean it's a work in progress or it just specifies what the method onFragmentInteraction should do? I also noticed there are three public interfaces with this same name on different classes of the code you sent me. Why is that? Could it be a good idea to write on another file and implement the interface in the classes? Maybe there's something i'm missing though.

BalzPit (commented):
The code is well written and easy to read. There are a few things I could't quite understand but it's probably because I didn't have the complete view of all the classes and their functions, maybe next time write a small summary of the general function of the class in the descriprion of the pull request. Being a newbie and having used Fragments in my code I also learned how to better and more safely create one.

forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
TaulantBullaku: thanks for the clarification, I will comment in more detail. thank you

PR #131, opened 2019-01-12
Codereview#2 of 5 (Bullaku) from 13 (Pham)

TaulantBullaku:
hello, I kindly ask you if you can do the codereview of my code inside ForumLogin.java


forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * String used to indicate that no username has been passed
+     */
+    private final String LOG_NULL_USERNAME = "A NULL username passed through liveLoginResponse";
+
+    /**
+     * String used to indicate that no error message has been passed
+     */
+    private final String LOG_NULL_ERROR_MESSAGE = "A NULL error message passed through liveError";
```
frankplus, line 51: Maybe you did this to avoid hard coding, but I think that this is a little bit excessive and unnecessary for this reason: these are just log messages for the programmer and it is not visualized in any way by the user, moreover these strings are used just once in the code. It would be easier to put the message directly in the log, this way we can avoid going up and down reading the code.

frankplus:
Hi Bullaku, in my opinion the code is easy to understand and sufficiently documented. Moreover I think it was a very good idea to take advantage of the model-view-controller design pattern, I took the opportunity to learn from this, thank you.

forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
TaulantBullaku: Yes, I did it to avoid hard coding.
I preferred doing this because i thought it was more important for the purpose of cleaner and more precise code.
But thanks for the advice

PR #132, opened 2019-01-14
CodeReview#2 of 16 (Perali) from 10 (Balzan)

Fisher4537:
Hi Pietro! I apologize for my delay.
Here is the code for the second review of the course.
I've developed the base of an Action Bar Fragment to improve the activity coordination.
This allow the user to have a context while switching between activity.
This Action Bar is implemented with a Fregment which can be added in each required activity writing only three lines of code. I think we need it for our app.
Could you please take a look also at the xml files that describe the action bar layout?
Thanks for your time!

P.s. This is not implemented in the common demo branch because we didn't have the activity for the user context there, however after the review we are going to design the finally app appearence.


app/src/main/java/unipd/se18/ocrcamera/ActionBarFragment.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * set the onClickListener for each buttons in the action_bar_layout
+     */
+    @Override
+    public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
+                             Bundle savedInstanceState) {
+
+        final View view = inflater.inflate(R.layout.action_bar_layout, container, false);
+
```
BalzPit: When you're dealing with fragments it's better to just return the inflated View in the onCreateView method, and then override onViewCreated() where you can initialize all layout elemets. Sometimes you'll need to call certain methods (such as getActivity() to get the context of the activity, which happens here in "mayStartActivity") that are unsafe to call in onCreateView.

app/src/main/java/unipd/se18/ocrcamera/ActionBarFragment.java
```diff
@@ -0,0 +1,92 @@
...
+public class ActionBarFragment extends Fragment {
+
+    /**
+     * set the onClickListener for each buttons in the action_bar_layout
+     */
+    @Override
+    public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
+                             Bundle savedInstanceState) {
+
+        final View view = inflater.inflate(R.layout.action_bar_layout, container, false);
```
BalzPit: When you're dealing with fragments it's better to just return the inflated View in the onCreateView method, and then override onViewCreated() where you can initialize all layout elemets. Sometimes you'll need to call certain methods (such as getActivity() to get the context of the activity, which happens here in "mayStartActivity") that are unsafe to call in onCreateView.

app/src/main/java/unipd/se18/ocrcamera/ActionBarFragment.java
```diff
@@ -0,0 +1,92 @@
...
+            public void onClick(View v) {
+                mayStartActivity(GalleryActivity.class);
+            }
+        });
+
+        ImageButton settingButton = (ImageButton)view.findViewById(R.id.action_bar_settings_button);
+        settingButton.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                mayStartActivity(SettingsActivity.class);
```
BalzPit, line 62: I like the modularity in this section, good job!

app/src/main/java/unipd/se18/ocrcamera/ActionBarFragment.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * if idSelectedButton equals one of the actionBarViewGroup children's id, the button
+     * state_selected is set to true. All the other children's state_selected are set to false.
+     * @param actionBarViewGroup is the view which contain the action buttons
+     * @param idSelectedButton is the id of the action button passed from R.id.button_id
+     */
+    protected static void enphasizeButton(ViewGroup actionBarViewGroup, int idSelectedButton) {
+        int totalActionButton = actionBarViewGroup.getChildCount();
```
BalzPit, line 92: The code here is good, but maybe add a comment to make it even easier to read.

app/src/main/java/unipd/se18/ocrcamera/ActionBarFragment.java
```diff
@@ -0,0 +1,92 @@
...
+        for (int i = 0; i < totalActionButton; i++) {
+            View currentButton = actionBarViewGroup.getChildAt(i);
+            if (currentButton.getId() == idSelectedButton) {
+                currentButton.setSelected(true);
+            } else {
+                currentButton.setSelected(false);
+            }
+        }
+    }
```
BalzPit, line 105: I thought that it might be a good idea to call this method inside the onClicks of the listeners inside this class so that everything that has to do with managing those buttons' states and looks is done in one place (maybe the method would need some changes but if i understand correctly the code it should not be that hard).
Maybe even better would be calling it in mayStartActivity after the second 'if', What do you think?

app/src/main/java/unipd/se18/ocrcamera/CameraActivity.java
```diff
@@ -1,13 +1,31 @@
...
+        ActionBarFragment.enphasizeButton(linearLayout, R.id.action_bar_camera_button);
+    }
+
+    @Override
+    protected void onPause() {
+        String msg = "onPause:";
+        Log.d(LOG_TAG, msg);
+        super.onPause();
     }
 }
```
BalzPit, line 31: I guess it's just an example of how to use enphasizeButton, but the part you wrote for this class seems fine to me, as does GalleryActivity and SettingsActivity. But there is something i don't understand: I might have totally missed it, but where is the Fragment created? Is it enough to just declare the fragment in the xml file of the activity?

app/src/main/res/layout/settings_activity.xml
```diff
@@ -0,0 +1,23 @@
...
+        android:layout_height="wrap_content"
+        android:layout_marginStart="8dp"
+        android:layout_marginTop="108dp"
+        android:layout_marginEnd="8dp"
+        android:text="@string/settingsTextView"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintHorizontal_bias="0.501"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent" />
+</android.support.constraint.ConstraintLayout>
```
BalzPit, line 23: Since I'm not an expert I couldn't really find any flaws in the xml files, I can only tell you that I could't have written them better than you. Good job nonetheless.

BalzPit (commented):
The code is pretty easy to understand and well moduled, there are just a couple of things i'm not really sure of, so i asked you in the comments.

Fisher4537:
Actually I'm thinking about how to emphasize the button from the ActionBarFragment! I can't do in the onClick method because after the click we want to start another activity and the state_selected of all the buttons is reset to false. The state_selected must change when the activity start!
Thanks to your suggestion I've implement the onViewCreated method. I've also added some comment in emphasizeButton().

PR #133, opened 2019-01-14
 Codereview#2 of 16 (Perali) from 4 (Pratesi)

Fisher4537:
Hi Leonardo! I apologize for my delay.
Here is the code for the second review of the course.
I've developed the base of an Action Bar Fragment to improve the activity coordination.
This allow the user to have a context while switching between activity.
This Action Bar is implemented with a Fregment which can be added in each required activity writing only three lines of code. I think we need it for our app.
Could you please take a look also at the xml files that describe the action bar layout?
Thanks for your time!

P.s. This is not implemented in the common demo branch because we didn't have the activity for the user context there, however after the review we are going to design the finally app appearence.


leopra:
I'm not understanding what is the purpose of having a Context while switching between activities, i think it would help me review the code

leopra:
I think that overall the code is clear

PR #136, opened 2019-01-14
Codereview#2 of 9 (Valente) from 6 (Furlan)

AlbVal97:
Hello @GiovanniFurlan ,
I kindly ask you to review part of the code I have sent to you.
To be specific, I need the classes ForumLogin.java and Login_VM.java to be reviewed, with the exception of the brief parts of Login_VM.java that only belong to @TaulantBullaku .
They are responsible for performing the login request to the database and then launch the fragments useful to access the forum content.

Thanks a lot for your time!
Alberto


forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
```diff
@@ -0,0 +1,182 @@
...
+ */
+public class ForumLogin extends Fragment {
+
+    /**
+     * ***************************
+     * **   GLOBAL VARIABLES    **
+     * ***************************
+     */
+
+    private OnFragmentInteractionListener mListener;
```
GiovanniFurlan, line 36: This variable is never initialized and use, i think u can delete it

forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
```diff
@@ -0,0 +1,182 @@
...
+            }
+        });
+
+    }
+
+    public interface OnFragmentInteractionListener
+    {
+        // TODO: Update argument type and name
+        void onFragmentInteraction(Uri uri);
+    }
```
GiovanniFurlan, line 181: I dont understand the meaning of this interface. Is that used somewhere? Does it needs to be completed? Otherwise i suggest u to delete it or at least complete the TODO.

forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
```diff
@@ -0,0 +1,182 @@
...
+}
```
GiovanniFurlan, line 182: I see some hardcoding, i suggest u to use the file strings.xml in resources to save your strings and then retrieve it when need. I see some case in line 41-46-51-56-169

forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
```diff
@@ -0,0 +1,182 @@
...
+        });
+
+        //When the login button is clicked a request is sent through the view model
+        loginButton.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+
+                //The username and password inserted are gathered from the EditText objects
+                userName = usernameEditText.getText().toString();
+                userPwd = pwdEditText.getText().toString();
```
GiovanniFurlan, line 160: I think that could be better to not let the password as a normal string in the code for security reason. i suggest u to cript it before use and send it to Login_VM. i know this is a base implementation so this is an advice for future improvement.

forum/src/main/java/unipd/se18/ocrcamera/forum/viewmodels/Login_VM.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * String used for debug logs to identify the view model throwing it
+     */
+    private final String LOG_TAG = "@@Login_VM";
+
+    /**
+     * String used to identify the error given by incorrect credentials
+     */
+    private final String LOG_INCORRECT_CREDENTIALS = "Connection established. Credentials refused.";
```
GiovanniFurlan, line 67: Variables in line 62 and 67 is never used, i think u forgot to delete them

forum/src/main/java/unipd/se18/ocrcamera/forum/viewmodels/Login_VM.java
```diff
@@ -0,0 +1,111 @@
...
+                    {
+                        forumLoginListener.onLoginFailure(context.getString(R.string.loginFailedMessage));
+                    }
+                }
+            }
+        };
+
+        //Sends a netword request to check if the provided credentials are correct
+        DatabaseManager.loginUser(context, username, password, listeners);
+    }
```
GiovanniFurlan, line 99: I see there are a lot of listeners, i suggest you to use different names to differentiate them. I had some difficulties to recognize them.

forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
AlbVal97: You're totally right: I'm going to implement this as soon as possible because, as you said, is a fairly basic security issue.

forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
AlbVal97: You're right: I think I forgot to delete it after the recent refactoring, so I'll get it done.

forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
AlbVal97: Due to the fact that it is linked to the first variable and it is never used (nor in @leopardo-rossi-developer class, which I used as an example) I think that I can safely delete it too.

forum/src/main/java/unipd/se18/ocrcamera/forum/ForumLogin.java
AlbVal97: Thank you so much for pointing them out! I definitely forgot to use resources for line 169 and I didn't even recognized as hardcoding the other four.

forum/src/main/java/unipd/se18/ocrcamera/forum/viewmodels/Login_VM.java
AlbVal97: That's true, I'm going to fix them. Thank you!

forum/src/main/java/unipd/se18/ocrcamera/forum/viewmodels/Login_VM.java
AlbVal97: You're right: I have replaced them with "forumLoginListeners" instead.

